PR #1228, opened 2026-04-20
fix(wish-state): auto-recover `completeGroup` from dispatch-bypass

### LINKED ISSUE #1214: bug: genie done silently no-ops when group isn't in_progress — entire wave progression breaks, engineers broadcast completion via prose instead
## Summary

`genie wish done <slug>#<group>` fails loudly (prints `❌ Cannot complete group "<N>": must be in_progress (currently ready)`, exits 1) when the group hasn't been transitioned to `in_progress` first. But from the orchestrator's vantage this looks like a **silent no-op**: the engineer's pane prints the error, the agent loop doesn't re-surface it, the agent broadcasts a prose completion message to team-lead, and team-lead's LLM interprets the prose as \"done\" and dispatches the next wave — all while the PG-backed `tasks` state never flips. The full 8-group `genie-serve-structured-observability` wish shipped and merged (PR #1213) with **zero** groups ever reaching `done` status in the state machine. Every dispatch was driven by teammate-message prose inference, not by the canonical completion signal.

## Why HIGH

`genie done` is the canonical completion signal for every engineer in every wish execution. When it effectively silent-fails, no wish auto-progresses through waves on the intended mechanism — execution only flows because LLMs infer progress from chat text. That path is unscriptable, unreproducible, and existentially fragile for autonomous wish execution.

## Reproduction Steps

1. Create a multi-group wish with dependencies (e.g. Group 1 → Group 2).
2. Spawn an engineer for Group 1 through ANY path that bypasses `runWorkDispatch` — direct `genie agent spawn`, SDK spawn, resume after a dispatch that failed at the `startGroup` step, or a dispatch flow where `startGroup` throws but the engineer was already booted.
3. Engineer completes the work, calls `genie done <slug>#1` (or `genie task done`/`genie wish done` — all routes converge on `doneCommand`).
4. Observe: engineer's pane prints `❌ Cannot complete group \"1\": must be in_progress (currently ready)` and exits 1.
5. Engineer's agent loop does not re-surface the CLI failure; it SendMessages team-lead with \"Group 1 complete\".
6. Run `genie wish status <slug>` — Group 1 still shows `ready`.
7. Team-lead's next-wave dispatch fires anyway because it reads prose, not state.

**Live evidence (2026-04-19, team `genie-serve-obs-v2`, wish `genie-serve-structured-observability`):**
- `23:02:35` — engineer-2 called `genie done genie-serve-structured-observability#2` — no state flip
- `23:02:42` — engineer-2 called `genie task done genie-serve-structured-observability#2` — no state flip
- `23:07:03` and later (including post-merge of PR #1213 on 2026-04-20) — `genie wish status` showed 0/8 done, all groups `ready`/`blocked`
- All 4 waves dispatched purely via teammate-message prose inference
- Full 8-group wish shipped and merged with 0 groups ever reaching `done` in PG

## Expected Behavior

1. When `genie done` is called on a group that isn't `in_progress`, either:
   - (a) Auto-recover: detect the dispatch-bypass, transition `ready → in_progress → done` atomically with a warning, OR
   - (b) Fail loudly AND propagate the failure to the agent orchestration layer so the teammate-message \"complete\" signal is not sent.
2. Teammate-message \"Group N complete\" broadcasts must NOT be emitted unless the `genie done` CLI exit was 0.
3. `genie wish status` should always reflect the authoritative state that `genie done` writes to.

## Actual Behavior

1. `completeGroup` at `src/lib/wish-state.ts:423-425` hard-gates on `child.status === 'in_progress'` and throws:
   ```ts
   if (child.status !== 'in_progress') {
     throw new Error(\`Cannot complete group \"\${groupName}\": must be in_progress (currently \${child.status})\`);
   }
   ```
2. `doneCommand` at `src/term-commands/state.ts:345-348` catches and exits:
   ```ts
   } catch (error) {
     const message = error instanceof Error ? error.message : String(error);
     console.error(\`❌ \${message}\`);
     process.exit(1);
   }
   ```
3. The engineer's shell sees the error. The engineer's **agent loop** does NOT inspect the exit code — it generates a prose \"Group N done\" teammate-message regardless, because the agent treats CLI output as advisory, not authoritative.
4. The next wave dispatches on prose. The state machine never advances. Compounded across 4 waves × 8 groups, the entire wish runs invisibly to the state system.

## Root Cause Analysis

**Investigation source:** `/trace` against the just-merged worktree `/home/genie/.genie/worktrees/genie/genie-serve-obs-v2` (branch `genie-serve-obs-v2`, PR #1213).

### Files & lines

| Role | File:Line | What it does |
|------|-----------|--------------|
| Write path entry | `src/term-commands/state.ts:303-350` | `doneCommand(ref)` — console route for `genie wish done` |
| Write path impl | `src/lib/wish-state.ts:413-466` | `completeGroup(slug, groupName)` — the only write |
| **Precondition gate** | `src/lib/wish-state.ts:423-425` | `if (child.status !== 'in_progress') throw` |
| Read path entry | `src/term-commands/state.ts:405-446` | `statusCommand(slug)` — console route for `genie wish status` |
| Read path impl | `src/lib/wish-state.ts:665-722` | `getState(slug)` — the only read |
| Ready→in_progress (single prod caller) | `src/term-commands/dispatch.ts:633` | `await wishState.startGroup(slug, group, agentName);` |
| Dispatch body | `src/term-commands/dispatch.ts:600-639` | `runWorkDispatch` — only path that calls `startGroup` |

### Causal chain

1. **Trigger:** Engineer runs `genie done <slug>#<group>`.
2. **Flow:** `doneCommand` → `wishState.completeGroup(slug, group)`.
3. **Fetch:** `completeGroup` calls `findGroup(sql, parent.id, groupName)` — succeeds, returns the child task row.
4. **Gate:** Line 423-425 evaluates `child.status !== 'in_progress'`. If the group was never transitioned to `in_progress` by `startGroup`, the current status is `ready` (or `blocked`), the gate fires, and the function throws.
5. **Catch:** `doneCommand` catches the throw at line 345-348, prints `❌ Cannot complete group \"<N>\": must be in_progress (currently ready)` to stderr, and `process.exit(1)`.
6. **Invisible to orchestrator:** The agent loop that scheduled the `genie done` call does NOT treat non-zero exits as work failure. It emits a prose \"Group N complete\" SendMessage to team-lead anyway.
7. **Team-lead infers progress:** Team-lead's LLM reads the prose, decides the wave is complete, dispatches next wave. State machine is not consulted.
8. **Observable symptom:** `genie wish status` always shows the original `ready`/`blocked` state. From outside, `genie done` looks like a silent no-op.

### Why dispatch-bypass happens in practice

`runWorkDispatch` (`dispatch.ts:600-639`) is the only production caller of `startGroup`. If engineers are spawned through ANY other path, the group never transitions. Known bypass vectors observed in this session:
- `genie agent spawn engineer` (direct spawn skips state machine — punch list bug D)
- Resume after `startGroup` threw at `dispatch.ts:635-639` with `process.exit(1)` but the engineer was booted separately
- SDK spawn paths that don't go through `dispatch.ts`

**There is no invariant check that reconciles \"engineer is running on group X\" with \"group X is `in_progress` in PG\".** The coupling is strictly forward (dispatch → startGroup → spawn) and has no enforcement against sidechannel spawns.

### Evidence the write and read agree on storage

To rule out the originally hypothesized \"doneCommand writes to different table than statusCommand reads from\":
- Both entry points import from the same module: `import * as wishState from '../lib/wish-state.js'` (`state.ts:18`).
- Both use `getConnection()` from `src/lib/db.js` (same DSN).
- Both query the PG `tasks` table keyed on `wish_file = .genie/wishes/<slug>/WISH.md` + `repo_path = resolveRepoPath(cwd)` + `parent_id IS NULL` (for parent) and `parent_id + group_name` (for child).
- There is no SQL table named `wish_groups` anywhere in the repo (`rg wish_groups --type sql` returns zero matches; the only textual hits are in-code comments).

The write/read SHOULD agree. They do agree — the failure is that `completeGroup` throws before writing.

### Confidence

**HIGH** for the precondition-gate root cause. The throw path is plainly coded and observable by running `genie done` on any `ready` group directly. The compound failure (engineer agent + team-lead prose inference) is what makes this look silent rather than loud.

### Tier 1 — make the failure loud at the orchestration layer (small, surgical)

1. **Agent loop must check CLI exit codes.** When an engineer runs `genie done` and the process exits non-zero, that's a hard failure. The agent MUST NOT emit a prose \"Group N complete\" teammate-message unless `genie done` exited 0.
2. **Team-lead must stop inferring progress from prose.** Next-wave dispatch must be gated on `genie wish status` returning `done` for all groups in the completing wave. Prose is advisory only.

### Tier 2 — close the dispatch-bypass window (architectural)

1. **Auto-recover in `completeGroup`:** if `child.status === 'ready'` and an active executor is tagged to this group, auto-transition `ready → in_progress → done` with a structured warning event emitted to the new observability stream (from PR #1213 — `wish_dispatch_bypass_recovered`).
2. **Add a DB trigger or application invariant:** any row in `executor_assignments` (or equivalent) pointing at `(wish_slug, group_name)` where the task is `ready`/`blocked` emits a `wish_state_invariant_violated` event. Background reconciler can choose to flip it or to kill the orphan executor.
3. **Unify spawn paths:** `genie agent spawn engineer --assign <wish>#<group>` must go through `runWorkDispatch`. Direct spawns without `--assign` should NOT carry implicit wish association; if they call `genie done`, the gate should fail with a different, specific error (`\"direct-spawn engineer cannot complete a wish group — use genie work instead\"`).

Tier 1 fixes the immediate operational problem (invisible failure). Tier 2 closes the underlying bypass. Both should land — Tier 1 first so the symptom stops compounding.

### Active Executors column drift (same family)

`printWishExecutors` at `state.ts:373-402` filters agents by `process.env.GENIE_TEAM`, not by the wish slug being statused. So when Felipe runs `genie wish status <slug>` from a shell without `GENIE_TEAM` set (or set to a different team), the column returns unrelated agents (his own session, trace subagent, etc.) instead of the engineers working the wish. This is a separate query-keying mistake with the same root pattern: **the read filter key doesn't match the write filter key.**

### No `wish_groups` SQL table

Original hypothesis was \"`genie done` writes to `wish_groups` but `genie wish status` reads from `tasks`\". Ruled out: no such table exists. The write/read both hit `tasks` + `task_dependencies` + `task_actors`.

## Environment

- **OS:** Linux (ProxmoxVE host, CT 171)
- **Runtime:** bun
- **Genie version:** v4.260419.4 (worktree HEAD `73ebfe28` on `genie-serve-obs-v2`, merged as PR #1213)
- **PG:** postgres://genie@localhost/genie
- **Repro context:** team `genie-serve-obs-v2` on wish `genie-serve-structured-observability` (8 groups, 4 waves, ALL groups shipped through prose-driven dispatch, 0 groups ever reached `done` in state machine)

## Evidence Sources

- `/trace` cascade against `src/term-commands/state.ts`, `src/lib/wish-state.ts`, `src/term-commands/dispatch.ts`
- BUGLESS GENIE punch list (2026-04-19), section #1 — HIGH
- PR #1213 merge diff for the branch where this was observed end-to-end

---

*Generated by genie `/report` on 2026-04-20. Part of the BUGLESS GENIE initiative to make genie self-healing by closing state-drift/silent-failure paths.*

namastex888:
## Summary
Fixes #1214 (Tier 1). `genie done <slug>#<group>` used to silent-fail when the group was in `ready` state (engineer spawned outside `runWorkDispatch`). The CLI exited 1, but the agent loop doesn't inspect exit codes — it broadcast "Group N complete" as prose regardless. Team-lead inferred progress from the prose and dispatched the next wave. An 8-group wish shipped with 0 groups reaching `done` in PG (PR #1213).

## Behavior changes in `completeGroup`
| Prior status | Before | After |
|--------------|--------|-------|
| `ready` | Throw `must be in_progress (currently ready)` | Auto-transition ready → in_progress (with `GENIE_AGENT_NAME` as assignee, warn to stderr), then complete |
| `done` | Throw `must be in_progress (currently done)` | Return existing done state (idempotent) |
| `blocked` | Throw `must be in_progress (currently blocked)` | Throw with named blockers: `blocked on unmet dependencies: <group> (<status>), ...` |
| `in_progress` | Complete | Complete (unchanged) |

## Scope
- Tier 1 only: CLI-layer resilience so `genie done` stops being a silent no-op.
- Tier 2 (agent loop gating teammate-message "complete" broadcasts on exit code; unified spawn path) remains on the issue.

## Test plan
- [x] `bun run check` — 3251 pass / 0 fail (pre-push gate green)
- [x] `bun test src/lib/wish-state.test.ts` — 76 pass; updated contract tests pin each new behavior
- [ ] Manual: dispatch-bypass reproduction in next real wish execution

## Notes
Cherry-picks #1224 (mock passthrough) so the pre-push gate passes — absorbed when #1224 merges.

Closes #1214

### DIFF
--- a/src/lib/protocol-router.test.ts
+++ b/src/lib/protocol-router.test.ts
@@ -46,6 +46,16 @@ mock.module('./tmux-wrapper.js', () => ({
   },
   genieTmuxPrefix: () => ['-L', 'genie', '-f', '/dev/null'],
   genieTmuxCmd: (sub: string) => `tmux -L genie ${sub}`,
+  // Passthrough matches the real implementation (issue #1223): the mock
+  // must preserve behavior because Bun's mock.module is process-global,
+  // so tmux-wrapper.test.ts can race and see this stub.
+  prependEnvVars: (command: string, env?: Record<string, string>) => {
+    if (!env || Object.keys(env).length === 0) return command;
+    const envArgs = Object.entries(env)
+      .map(([k, v]) => `${k}=${v}`)
+      .join(' ');
+    return `env ${envArgs} ${command}`;
+  },
 }));
 
 mock.module('./orchestrator/index.js', () => ({
--- a/src/lib/spawn-command.test.ts
+++ b/src/lib/spawn-command.test.ts
@@ -214,6 +214,16 @@ mock.module('./tmux-wrapper.js', () => ({
   },
   genieTmuxPrefix: () => ['-L', 'genie', '-f', '/dev/null'],
   genieTmuxCmd: (sub: string) => `tmux -L genie ${sub}`,
+  // Passthrough matches the real implementation (issue #1223): the mock
+  // must preserve behavior because Bun's mock.module is process-global,
+  // so tmux-wrapper.test.ts can race and see this stub.
+  prependEnvVars: (command: string, env?: Record<string, string>) => {
+    if (!env || Object.keys(env).length === 0) return command;
+    const envArgs = Object.entries(env)
+      .map(([k, v]) => `${k}=${v}`)
+      .join(' ');
+    return `env ${envArgs} ${command}`;
+  },
 }));
 
 mock.module('./orchestrator/index.js', () => ({
--- a/src/lib/tmux-alive.test.ts
+++ b/src/lib/tmux-alive.test.ts
@@ -10,6 +10,16 @@ mock.module('./tmux-wrapper.js', () => ({
   executeTmux: mockExecuteTmux,
   genieTmuxPrefix: () => ['-L', 'genie', '-f', '/dev/null'],
   genieTmuxCmd: (sub: string) => `tmux -L genie ${sub}`,
+  // Passthrough matches the real implementation (issue #1223): the mock
+  // must preserve behavior because Bun's mock.module is process-global,
+  // so tmux-wrapper.test.ts can race and see this stub.
+  prependEnvVars: (command: string, env?: Record<string, string>) => {
+    if (!env || Object.keys(env).length === 0) return command;
+    const envArgs = Object.entries(env)
+      .map(([k, v]) => `${k}=${v}`)
+      .join(' ');
+    return `env ${envArgs} ${command}`;
+  },
 }));
 
 const { isPaneAlive, isPaneProcessRunning, TmuxUnreachableError } = await import('./tmux.js');
--- a/src/lib/tmux-resolve.test.ts
+++ b/src/lib/tmux-resolve.test.ts
@@ -14,6 +14,16 @@ mock.module('./tmux-wrapper.js', () => ({
   executeTmux: mockExecuteTmux,
   genieTmuxPrefix: () => ['-L', 'genie', '-f', '/dev/null'],
   genieTmuxCmd: (sub: string) => `tmux -L genie ${sub}`,
+  // Passthrough matches the real implementation (issue #1223): the mock
+  // must preserve behavior because Bun's mock.module is process-global,
+  // so tmux-wrapper.test.ts can race and see this stub.
+  prependEnvVars: (command: string, env?: Record<string, string>) => {
+    if (!env || Object.keys(env).length === 0) return command;
+    const envArgs = Object.entries(env)
+      .map(([k, v]) => `${k}=${v}`)
+      .join(' ');
+    return `env ${envArgs} ${command}`;
+  },
 }));
 
 const { resolveRepoSession } = await import('./tmux.js');
--- a/src/lib/tmux.test.ts
+++ b/src/lib/tmux.test.ts
@@ -12,6 +12,16 @@ mock.module('./tmux-wrapper.js', () => ({
   executeTmux: mockExecuteTmux,
   genieTmuxPrefix: () => ['-L', 'genie', '-f', '/dev/null'],
   genieTmuxCmd: (sub: string) => `tmux -L genie ${sub}`,
+  // Passthrough matches the real implementation (issue #1223): the mock
+  // must preserve behavior because Bun's mock.module is process-global,
+  // so tmux-wrapper.test.ts can race and see this stub.
+  prependEnvVars: (command: string, env?: Record<string, string>) => {
+    if (!env || Object.keys(env).length === 0) return command;
+    const envArgs = Object.entries(env)
+      .map(([k, v]) => `${k}=${v}`)
+      .join(' ');
+    return `env ${envArgs} ${command}`;
+  },
 }));
 
 // Must import after mock.module
--- a/src/lib/wish-state.test.ts
+++ b/src/lib/wish-state.test.ts
@@ -208,14 +208,18 @@ describe.skipIf(!DB_AVAILABLE)('pg', () => {
       expect(state?.groups['4'].status).toBe('ready');
     });
 
-    test('refuses when group already done', async () => {
+    test('idempotent when group already done (issue #1214)', async () => {
+      // Previously threw `must be in_progress (currently done)`. That caused
+      // `genie done` to exit 1 when an engineer called it twice (e.g. retry),
+      // which the orchestrator couldn't distinguish from a real failure.
+      // Now: return the existing done state without throwing.
       await createState('test-wish', sampleGroups, cwd);
       await startGroup('test-wish', '1', 'agent-a', cwd);
       await completeGroup('test-wish', '1', cwd);
 
-      await expect(completeGroup('test-wish', '1', cwd)).rejects.toThrow(
-        'Cannot complete group "1": must be in_progress (currently done)',
-      );
+      const result = await completeGroup('test-wish', '1', cwd);
+      expect(result.status).toBe('done');
+      expect(result.completedAt).toBeDefined();
     });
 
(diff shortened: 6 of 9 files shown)


namastex888:
Addressed review feedback in 2b06932a:

**Codex P2 (wish-state.ts:482) — FIXED**
You're right, `child.started_at` stayed undefined after the PG UPDATE, so `toISO(child.started_at)` at the return builder leaked `startedAt: undefined` to any caller using GroupState for timing. Mirrored the PG `COALESCE(started_at, $1)` in memory: `if (!child.started_at) child.started_at = startNow;`. Added a `startedAt` assertion to the auto-recovery test so this never regresses silently.

**Gemini MEDIUM (wish-state.ts:438) — FIXED**
Parallelized the two idempotent-done queries with `Promise.all`. Independent reads, no dependency — small but real win.

**Gemini MEDIUM (protocol-router.test.ts:58) — DEFERRED**
Style-only DRY suggestion. The `createTmuxWrapperMock` helper is a nice-to-have but the duplication is from merged #1224 and out of scope for this PR. Happy to file as a follow-up if you think it's worth it.

All 3251 tests still green.

PR #1413, opened 2026-04-27
fix(spawn+ls): ghost tmux sessions + bare-name shadow rendering — §19 v2 1:1 invariant

namastex888:
Two surfaces of the same architectural violation: **`session_name = owning_agent` strictly 1:1** (per §19 v2). Tonight's master-aware-spawn Wave 2 dispatch broke this in both the spawn write-path AND the read-path.

## The bug Felipe caught

I (`genie@genie`, attached to tmux session `genie`) spawned three engineers + a fixer with `--cwd <worktree> --new-window`. Expected them to land in MY session as new windows. Reality:

```
$ tmux list-sessions
$11 @55  windows=4   ← Wave 2 engineers (engineer-w2g4/7/14)
$12 @60  windows=2   ← fix-shadow-render
$5  genie windows=4  ← my expected session
```

Workers existed in PG with `state='running'`, valid pids, growing jsonls. From Felipe's POV: **ghost windows**. The orphan sessions had no human-readable name, never appeared in his TUI, and when the workers died (pre-push hook flake → process exit → orphan-session collapse), no death-hook updated PG. Stale `running` rows shadowed the actual idle UUID-keyed peers in `genie ls` and `genie status`.

### 1. Spawn write-path (`src/lib/protocol-router-spawn.ts:293`) — §19 v2 invariant

Before:
```typescript
const session =
  teamConfig?.tmuxSessionName ?? (await resolveRepoSession(repoPath))
    ?? (await getCurrentSessionName()) ?? team;
```
For `--cwd <worktree>`, `resolveRepoSession` derived the worktree basename, didn't find a matching tmux session, returned the basename verbatim. `ensureTeamWindow` called `tmux new-session` with it but tmux's auto-id naming kicked in and the session ended up `@55`/`@60`. Even when the name landed correctly, this violates the 1:1 rule: **hires are children of the spawning leader; they MUST inhabit the leader's session**.

After:
```typescript
const session = (await getCurrentSessionName()) ?? teamConfig?.tmuxSessionName ?? team;
```
- Spawner's CURRENT session wins first (the team-leader's own session by 1:1 rule).
- `resolveRepoSession(repoPath)` is **removed from the hot path** — `repoPath` is the worker's working directory; it MUST NOT dictate which tmux session hosts the worker.
- `resolveRepoSession` remains in legitimate callers (`term-commands/team.ts` for team-create, `term-commands/agents.ts` for `register`), where repo→session inference is the explicit intent.

### 2. Read-path bare-name shadow dedup (`8167be59` — already in this branch)

`registry.list()` returned both the bare-name row (legacy, no executor) and the UUID-keyed live row for the same `custom_name`. Renderers (`genie ls` + `genie status`) treated them as separate agents and surfaced the bare-name row's stale signals — making LIVE workers appear `idle` / `auto_resume_disabled` / `no-session`.

Fix adds `registry.listForRender()` + 5 tests covering: shadow→live coalescing, bare-name-alone preserved, UUID-alone preserved, dedup is per-`custom_name` only.

## Commits

| SHA | Summary |
|-----|---------|
| `8167be59` | `fix(read-path): dedupe bare-name shadow rows in agent listings — surface live signals` |
| `<head>`   | `fix(spawn): hire workers into spawner's tmux session — §19 v2 1:1 rule` |

## Validation

- `bun run typecheck` clean
- `bun test src/lib/protocol-router-spawn.test.ts` 5 pass / 0 fail (regression guard for the spawn fix)
- `bun test src/lib/agent-registry.test.ts` (read-path dedup tests pass per the fixer's earlier run)
- `resolveRepoSession` still imported + used by 2 legitimate callers (verified via grep)

## Why one PR

Both fixes target the same architectural invariant — that `session_name = owning_agent_name` strictly 1:1. The write-path fix prevents ghost sessions on FUTURE spawns; the read-path fix surfaces correct state for any agent whose lifecycle already produced a shadow. Shipping together gives Felipe end-to-end coverage of the "where is my worker?" UX gap.

## What's NOT in this PR

- The Wave 2 master-aware-spawn-w2 engineers' actual code changes (Group 4 partition self-heal, Group 7 jsonl team-relax, Group 14 master backfill). They committed locally before dying but never pushed. We'll respawn them in a fresh worktree AFTER this PR lands so their next attempt uses the §19-corrected spawn-path.
- Stale-executor reconciler that detects orphan-session collapse and updates `executors.state='terminated'`. Separate concern (matches Wave 1's Group 10 stale-spawning reaper).

### DIFF
--- a/src/lib/agent-registry.test.ts
+++ b/src/lib/agent-registry.test.ts
@@ -16,6 +16,8 @@ import {
   getTeamLeadEntry,
   list,
   listAgents,
+  listAgentsForRender,
+  listForRender,
   listTemplates,
   reconcileStaleSpawns,
   register,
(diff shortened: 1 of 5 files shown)


namastex888:
Addressed Codex P1 + rebased on dev (`121074e0`).

**Codex P1 (RESTORED team-config precedence over inferred current session):**
The session-resolution chain now gates step-1's `getCurrentSessionName()` on actual `process.env.TMUX` attachment:

```typescript
const inTmux = Boolean(process.env.TMUX);
const callerSession = inTmux ? await getCurrentSessionName() : null;
const session = callerSession ?? teamConfig?.tmuxSessionName ?? team;
```

Resolution order:
1. **Caller is attached to tmux** (`TMUX` env set) → use their session. This is the §19 v2 1:1 case (interactive leader hires worker → worker lands in leader's session).
2. **Caller is NOT in tmux** (CLI/SDK/background) → use the team's configured `tmuxSessionName`.
3. **Last resort** → literal team name.

This avoids the regression Codex flagged: `getCurrentSessionName()` falling back to "first session from `tmux list-sessions`" when `TMUX` is unset, which would silently route background/SDK spawns into an unrelated session and ignore `teamConfig.tmuxSessionName`.

**Gemini's unification suggestion** (extract shared helper, unify with `agents.ts:resolveSpawnTeamWindow`) is good but deferred to a follow-up to keep this PR focused on the live ghost-window bug. Filing as a Wave 3 punch-list item.

**The PG Tests failure** on the previous run was the `CONNECTION_ENDED 127.0.0.1:20900` pgserve port-flake we've been hitting all evening (`pg > register and get > register creates agent, get retrieves it` fails first because it's first to acquire the connection on cold-start). Not related to this diff. Hoping the rebase + retry clears it.

PR #1507, opened 2026-04-29
fix(tui): zero-counter regression in workspace mode (#1447 follow-up)

namastex888:
## Summary

Fixes the `0/0` nav-header regression introduced by #1447. `computeNavCounts` returned zero whenever `workspaceRoot` was set but the workspace scan came up empty (stale `workspaceRoot`, no agents resolved, or every node was an orphan session) — even when tmux had live panes.

- Falls back to live tmux pane inventory when workspace mode produces zero agents
- Exports `computeNavCounts` for unit-test coverage — the missing test on this helper is exactly why the regression slipped through #1447's review
- Adds 8 regression tests: workspaceRoot+agents, workspaceRoot+empty fallback, no-workspaceRoot tmux sum, mixed orphan/agent counting, running filter, diagnostic-snapshot edges

## Test plan

- [x] `bun test src/tui/components/Nav.test.tsx` — 8 new tests pass
- [ ] Manual: launch with stale `workspaceRoot` + live tmux panes → header shows pane count instead of 0/0

### DIFF
--- a/src/tui/components/Nav.test.tsx
+++ b/src/tui/components/Nav.test.tsx
@@ -11,8 +11,9 @@
  */
 
 import { describe, expect, mock, test } from 'bun:test';
+import { buildWorkspaceTree } from '../session-tree.js';
 import type { TreeNode } from '../types.js';
-import { handleEnterAgent } from './Nav.js';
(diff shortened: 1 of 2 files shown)


namastex888:
Tracked by #1509 (orphan-recovery sprint).

PR #1530, opened 2026-04-29
fix(scheduler): W1 hotfix — move resume.missing_session emission from read path to scheduler

### LINKED ISSUE #1461: Bug: shouldResume status reads emit unbounded audit events on every TUI tick
## Summary

The `getResumeSessionId` function (`src/lib/executor-registry.ts:516`) is documented as a "single-reader chokepoint" called from many surfaces — TUI badge rendering, `genie ls`, `genie status`, scheduler decisions, spawn paths. Every call to it on an orphan agent row (`current_executor_id IS NULL`) writes a `resume.missing_session` audit event with no de-dupe and no rate-limit. Combined with the TUI's 2s polling tick, this turns a status read into an unbounded audit-event firehose for any agent whose executor row has been nullified.

Verified against HEAD `0c3ed416` (origin/dev).

## Reproduction

1. Insert an `agents` row with `auto_resume=true`, `current_executor_id=NULL`, no matching JSONL on disk:
   ```sql
   INSERT INTO agents (id, ..., auto_resume, current_executor_id) VALUES ('test-orphan', ..., true, NULL);
   ```
2. Open the genie TUI.
3. Wait ~10 seconds.
4. Query the audit log:
   ```bash
   genie events list --entity test-orphan --since 1m
   ```
5. Observe `resume.missing_session reason=no_executor` events accumulating at ~1 per Nav refresh tick (every 2s; faster during spawn-ready watching at 250ms).

In the wild this surfaced as ~10 events/s in the PG events table for a single orphan row, indefinitely.

## Call chain (verified at HEAD `0c3ed416`)

\`\`\`
src/tui/components/Nav.tsx:57           setInterval(refresh, 2000)  ← loop driver
  → src/tui/diagnostics.ts:241          collectDiagnostics()
    → src/tui/db.ts:103                 shouldResume(a.id) for every non-archived agent
      → src/lib/should-resume.ts:189    getResumeSessionId(agentId)
        → src/lib/executor-registry.ts:549   emitMissingSession(agentId, row, actor)
          → src/lib/executor-registry.ts:505 recordAuditEvent('resume.missing_session', { reason: 'no_executor' })
\`\`\`

Other callers of the same chokepoint (each one a potential write site for a status read):

- \`src/term-commands/status.ts:88\`  — \`genie status\`
- \`src/term-commands/agents.ts:1635, 2696, 2839, 2873, 3189\` — multiple agent-listing surfaces
- \`src/lib/team-auto-spawn.ts:194\` — leader-resume probe
- \`src/lib/protocol-router.ts:317\` — protocol-router decision
- \`src/lib/protocol-router-spawn.ts:30\` (import; calls in body)
- \`src/lib/scheduler-daemon.ts\` (multiple references in resume flow)

## Why this matters

- **Audit-log noise** — operator filters drown in retry-tick events; real signals get buried.
- **DB write amplification** — every TUI session × every orphan agent × every 2s tick. With N orphans + M open TUIs, this is N×M×0.5 writes/s into \`audit_events\` for zero new information.
- **Hides rather than surfaces a real problem** — the orphan row's existence is what should be flagged once, not the same condition restated 30,000 times an hour.

The chokepoint pattern is good — having one place to ask "should this agent resume?" is the right design. The flaw is that the chokepoint silently writes to the audit log on every read.

## Repair shape (one of)

**Option A (least-invasive) — silent read mode.** Add a \`silent: true\` parameter to \`getResumeSessionId\` that skips \`emitMissingSession\`/\`recordAuditEvent\`. TUI badge path passes \`silent: true\`; scheduler/spawn paths that genuinely *act* on the decision keep emitting. Gates on intent: status reads stay quiet, decision points stay observable.

\`\`\`ts
// src/lib/executor-registry.ts:516
export async function getResumeSessionId(
  agentId: string,
  opts: { silent?: boolean } = {},
): Promise<string | null>
\`\`\`

Call sites that should pass \`silent: true\`: \`tui/db.ts:103\`, \`term-commands/status.ts:88\`, \`term-commands/agents.ts\` listing surfaces, \`team-auto-spawn.ts:194\` (probe), \`protocol-router.ts:317\` (probe).

Call sites that should keep emit (decision points): \`scheduler-daemon.ts\` resume flow, \`term-commands/agents.ts:2839\` (the actual recover path).

**Option B — split read/decide.** Rename current function to \`getResumeSessionIdAndAudit()\`; introduce \`peekResumeSessionId()\` that does the same query but no emit. Status callers use \`peek\`, decision callers use \`AndAudit\`. Better signal-to-noise in code review (the function name tells you what it does), but more sites to update.

**Option C — dedupe by (agentId, reason) within a window.** Cache last-emit timestamp per (agentId, reason) and skip if within e.g. 60s. Simplest one-liner, but suppresses legitimate decision-point emits during the window. Not recommended.

Recommended: **Option A**. Smallest diff, preserves observability where it matters, kills the firehose where it doesn't.

## Reference

This was found while investigating a runaway resume loop on a user-attached Claude Code session (\`genie-pgserve\`) whose registration left an orphan \`current_executor_id=NULL\`. The TUI fired \`resume.missing_session\` >10 times/sec without ever advancing state. Issues #1462 (counter never trips) and #1463 (auto_resume default) compound the problem; this issue is the loop's tight inner cycle.

namastex888:
## Summary

**Hotfix for PR #1528.** dog-fooder-da66 verdict 2026-04-29 verified that PR #1528's in-process dedupe was insufficient for the production storm pattern: synthetic orphan still emitted **368 cli + 66 genie events in 67s** because each `genie ls` / TUI tick / status read is a fresh process with no shared in-memory state.

This PR takes the architecturally correct approach: emission belongs to the actor-that-attempts-the-action (scheduler), not the actor-that-queries-state (CLI/TUI/protocol-router). The read path now emits **zero** events; the scheduler emits one per orphan per tick (bounded by the 60s daemon cadence).

### Read path (`src/lib/executor-registry.ts`)
- `getResumeSessionId` no longer emits `resume.missing_session`
- `emitMissingSession()` removed; replaced by exported `recordResumeMissingSession(agentId, actor, details)` that the scheduler calls explicitly
- In-process dedupe map removed (load-bearing was illusory)
- `_resetMissingSessionDedupeForTests()` retained as a no-op for back-compat

### Scheduler (`src/lib/scheduler-daemon.ts attemptAgentResume`)
- In the `!currentSessionId` branch, after counter increment, calls `recordResumeMissingSession()` to emit one event per scheduler tick
- Worst-case: 1 event / orphan / 60s regardless of CLI process count

## Tests

| File | Existing | Modified | Added | Total | Status |
|------|---------:|---------:|------:|------:|--------|
| `executor-registry.test.ts` | 92 | 4 | 0 | 96 | ✅ |
| `auto-resume-zombie-cap.test.ts` | 18 | 0 | 2 | 20 | ✅ |
| `scheduler-daemon.test.ts` | 113 | 0 | 0 | 113 | ✅ no regression |
| `should-resume.test.ts` | 109 | 0 | 0 | 109 | ✅ no regression |
| `agent-registry.test.ts` | 93 | 0 | 0 | 93 | ✅ no regression |

Modified tests flip from "asserts emission happens" to "asserts NO emission from read path" — that's the new contract.

The 2 new `auto-resume-zombie-cap.test.ts` tests use a real pgserve so we can verify the audit row landed:
- `attemptAgentResume on orphan with no session emits exactly one resume.missing_session per call`
- `three sequential ticks produce three events (bounded by tick cadence, not unbounded)`

## Why PR #1528's dedupe failed

PR #1528 added an in-process Map keyed by `(agentId, actor, reason)` with a 60s TTL. That works inside one process. But:

- `genie ls` / `genie agent show` / TUI ticks each spawn a **new process**
- Each process starts with an empty Map, emits its first event
- 5+ such processes per second = 300+ events / minute / orphan
- Daemon-side reads (genie serve) have their own Map but were correctly deduping → da66 saw 66 daemon events vs 368 cli events

This hotfix removes the question entirely by moving the emission point to a long-running process.

## Validation

- All affected test files pass in isolation
- `bun run typecheck` clean
- Awaiting da66's verification on this branch

## Closes (re-opens + actually closes)

- Closes #1461 (in addition to PR #1528 which was VERIFIED-BROKEN)

## Related

- Builds on #1528 (PR #1528 — the original W1 attempt that introduced #1461/#1462/#1463)

### DIFF
--- a/src/lib/__tests__/auto-resume-zombie-cap.test.ts
+++ b/src/lib/__tests__/auto-resume-zombie-cap.test.ts
@@ -555,3 +555,97 @@ describe('#1462: orphan agents (no currentSessionId) increment counter and exhau
     expect(agentUpdates.length).toBe(0);
   });
 });
+
+// ---------------------------------------------------------------------------
+// W1 hotfix — `resume.missing_session` emission contract
+//
+// PR #1528 emitted from the read path (getResumeSessionId) with in-process
+// dedupe; that didn't survive CLI process boundaries (each `genie ls` is a
+// fresh process). PR #1530 (this) moves emission to the scheduler's
+// attemptAgentResume no-session branch — bounded to ≤1 emit per orphan per
+// scheduler tick (60s) by the daemon's tick rate.
+//
+// These tests use a real pgserve so we can verify audit rows landed
+// (createMockSql() returns [] for every query so audit writes are no-ops
+// there). We dispatch the actual function and read audit_events.
+// ---------------------------------------------------------------------------
+
+import { afterAll, beforeAll, beforeEach } from 'bun:test';
+import { findOrCreateAgent } from '../agent-registry.js';
+import { getConnection } from '../db.js';
+import { DB_AVAILABLE, setupTestDatabase } from '../test-db.js';
+
+describe.skipIf(!DB_AVAILABLE)('W1 hotfix: scheduler emits resume.missing_session for orphan path', () => {
+  let cleanupDb: () => Promise<void>;
+
+  beforeAll(async () => {
+    cleanupDb = await setupTestDatabase();
+  });
+
+  afterAll(async () => {
+    await cleanupDb();
+  });
+
+  beforeEach(async () => {
+    const sql = await getConnection();
+    await sql`DELETE FROM audit_events WHERE event_type = 'resume.missing_session'`;
+    await sql`DELETE FROM agents WHERE id LIKE 'w1-hotfix-%'`;
+  });
+
+  async function countMissingSessionEvents(agentId: string): Promise<number> {
+    const sql = await getConnection();
+    const rows = await sql<{ count: number }[]>`
+      SELECT COUNT(*)::int AS count
+      FROM audit_events
+      WHERE entity_type = 'agent'
+        AND entity_id = ${agentId}
+        AND event_type = 'resume.missing_session'
+    `;
+    return rows[0].count;
+  }
+
+  test('attemptAgentResume on orphan with no session emits exactly one resume.missing_session per call', async () => {
+    const agent = await findOrCreateAgent('w1-hotfix-orphan-1', 'w1-hotfix-team', 'engineer');
+
+    const orphan = makeWorker({
+      id: agent.id,
+      currentSessionId: null,
+      resumeAttempts: 0,
+      maxResumeAttempts: 3,
+      autoResume: true,
+    });
+
+    const { deps } = createMockDeps({
+      getConnection: async () => await getConnection(),
+    });
+
+    expect(await countMissingSessionEvents(agent.id)).toBe(0);
+    await attemptAgentResume(deps, defaultConfig, orphan);
+    expect(await countMissingSessionEvents(agent.id)).toBe(1);
+  });
+
+  test('three sequential ticks produce three events (bounded by tick cadence, not unbounded)', async () => {
+    const agent = await findOrCreateAgent('w1-hotfix-orphan-3', 'w1-hotfix-team', 'engineer');
+    const { deps } = createMockDeps({
+      getConnection: async () => await getConnection(),
+    });
+
+    // Three scheduler ticks back-to-back. In production these are 60s apart
+    // by daemon cadence; the test compresses time. Each tick increments the
+    // counter and emits one event until exhaustion at attempt=3.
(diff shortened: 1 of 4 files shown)


namastex888:
Closing to reopen with conventional-commit-compliant message (header was 106 chars vs 100 max + uppercase subject). Same code, fresh branch incoming.

PR #1548, opened 2026-04-29
fix(pgserve): start v2 daemon from serve

namastex888:
## Summary
- make socket-mode getConnection ensure pgserve daemon exists before dialing /tmp/pgserve
- make genie serve start the v2 daemon instead of legacy --port 19642 TCP pgserve
- keep Genie data at the existing ~/.genie/data/pgserve path and print it in startup output
- consume pgserve SDK when available, with CLI fallback for pgserve@2.0.0

## Tests
- bun test src/lib/db.test.ts src/term-commands/serve.test.ts
- bun test src/term-commands/serve.test.ts
- bun run typecheck
- bun run build
- bunx biome check src/lib/db.ts src/lib/db.test.ts src/term-commands/serve.ts src/term-commands/serve.test.ts src/types/pgserve.d.ts
- bun run lint (repo warnings only; command exits 0)
- pre-push gate ran during git push

### DIFF
--- a/src/lib/db.test.ts
+++ b/src/lib/db.test.ts
@@ -337,6 +337,29 @@ describe('daemon-owned pgserve', () => {
     // Self-heal function exists
     expect(source.includes('selfHealPostgres')).toBe(true);
   });
+
+  test('socket connections ensure the v2 daemon before dialing libpq path', () => {
+    const source = readFileSync(join(__dirname, 'db.ts'), 'utf-8');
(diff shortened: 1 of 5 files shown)


namastex888:
Closing as superseded.

Current dev has the safe v1/v2 coexistence path from #1549 and the socket-auth lockout fix from #1550. Rebasing this PR would either be empty or would reintroduce the older daemon behavior that shared legacy v1 state / stale pid cleanup, so it should not be merged.

Issue #1574, opened 2026-04-30
PG connection leak: short-lived CLI/hook subprocesses skip shutdown() — saturates max_connections


namastex888:
## Live observation (production, 2026-04-30 ~13:29 UTC)

After running `genie@4.260430.15` with `pgserve@2.0.8` for ~24h with 5+ active Claude/codex agents (each firing PreToolUse/PostToolUse hooks), the postgres backend serving genie's data dir saturated `max_connections=1000`.

**Hard evidence:**

```bash
$ psql -h /run/user/1000/pgserve -d postgres -c "SELECT 1"
psql: error: connection to server on socket "/run/user/1000/pgserve/.s.PGSQL.5432" failed:
FATAL:  sorry, too many clients already
```

```bash
$ ps -eo pid,ppid,cmd | awk '/postgres:/' | awk '{print $2}' | sort | uniq -c | sort -rn
   1008 2453094  ← genie's data dir postgres root, max_connections=1000 SATURATED
     18 2523378  ← omni's DB (normal-shape pool baseline)
```

Could not even open a debug `psql` session — every connection slot exhausted. System effectively DoS'd itself.

### File: `src/lib/db.ts:1159-1174`

```typescript
sqlClient = pgModule({
  host, port, database,
  username: DB_NAME,
  [PG_AUTH_FIELD]: pgWireCredential,
  max: 50,                  // ← 50 connections per subprocess
  idle_timeout: 1,
  connect_timeout: 5,
  onnotice: () => {},
  connection: { client_min_messages: 'warning' },
});
```

Per-subprocess pool of `max: 50`. The singleton `sqlClient` only de-dupes within a single Node process. **Each genie CLI invocation = separate Node process = its own pool of up to 50.**

### File: `src/hooks/dispatch-command.ts:41`

```typescript
process.exit(0);   // ← exits without shutdown()
```

**Zero `shutdown()` calls in `src/hooks/**`** (verified: `grep -rEn 'shutdown\(\)' src/hooks/` returns empty). Hooks fire on every Claude tool call across every agent in the system; each invocation spawns a subprocess that:
1. Loads `dist/genie.js`
2. Initializes module-level `sqlClient = null`
3. On first DB call → creates pool (max: 50)
4. Does 1-3 queries
5. **`process.exit(0)`** without graceful pool close

### Why connections leak server-side

[...]

PR #1580, opened 2026-04-30
fix(pgserve): pin cwd for stable app identity + drain pool on exit (#1575, #1574)

### LINKED ISSUE #1575: spawn/wish/task wedged: missing fingerprinted DB `app_anon_<fp>` not auto-created
## Summary

`genie spawn`, `genie wish status`, `genie task create`, and similar PG-backed commands fail with:

```
Error: database "app_anon_35da1b32060e" does not exist
```

The fingerprint (`35da1b32060e`) is the workspace-level genie state DB for `/home/genie/workspace/agents/khal-os`. It is **referenced** by the agent-directory / wish / task subsystems but **does not exist** in pgserve.

The only escape hatch I have as a user is `genie db query "CREATE DATABASE app_anon_<fp>"`, which (a) requires me to know the missing fingerprint and (b) leaves the schema un-initialized. Genie should detect a missing fingerprinted DB on first connect and either auto-create+migrate it or fail with an actionable hint.

## Reproduction

In a workspace whose fingerprinted DB does not yet exist (e.g. fresh agent team, or a workspace whose anon DB was pruned):

```bash
$ genie wish status some-slug
❌ database "app_anon_35da1b32060e" does not exist

$ genie spawn engineer --prompt "..."
[agent-directory] lookupTemplateTeam(engineer) failed: database "app_anon_35da1b32060e" does not exist
Error: database "app_anon_35da1b32060e" does not exist

$ genie task create "..." --type software
Error: database "app_anon_35da1b32060e" does not exist

$ genie db migrate
Error running migrations: database "app_anon_35da1b32060e" does not exist
```

`genie db ls` confirms the fingerprint is absent (~20 other `app_anon_*` DBs present, but not `35da1b32060e`).

`genie doctor` reports green — no diagnostic surface for this failure.

## Environment

- genie: `4.260430.15` (bun)
- workspace: `/home/genie/workspace/agents/khal-os` (khal-os meta-repo, multi-submodule)
- pgserve: connected (`genie db ls` works, `genie db url` works)
- `genie db status` reports "stopped" but `genie db ls` clearly works — also a bug, see "Related" below

## Impact

When this hits, the genie CLI's core orchestration verbs are unusable in the affected workspace. The user has to either:
- Manually `CREATE DATABASE` (which sandboxes correctly flag as shared-infra modification)
- Fall back to non-genie dispatch tooling

This breaks the `/wish` → `/work` skill chain because `/work` dispatches via `genie spawn` and `genie work`, and both fail.

## Expected Behavior

On first connect for a fingerprinted workspace whose DB does not exist:
1. Genie auto-creates `app_anon_<fp>` (or `app_<package>_<fp>` if `package.json` is present)
2. Runs migrations to bring schema up to current
3. Either succeeds silently or prints a one-line "initialized workspace DB" notice

If auto-create is undesirable (security/policy), the error should be actionable:

```
Error: workspace DB app_anon_35da1b32060e does not exist.
Fix: genie db init  (or: genie db query "CREATE DATABASE app_anon_35da1b32060e" && genie db migrate)
```

Currently the error is just the raw PG message with no remediation path.

## Related

- `genie db status` reports "stopped" while pgserve is clearly running and accepting connections (other `genie db` subcommands work). Likely a separate bug in the status probe.
- `genie doctor` should detect missing fingerprinted DB and warn — currently it doesn't.

## Workaround

```bash
genie db query "CREATE DATABASE app_anon_<missing-fingerprint>"
genie db migrate
```

This requires extracting the fingerprint from the error message, and may leave subsystems in an undefined state if migrate doesn't fully bootstrap a fresh anon DB.

namastex888:
## Summary

Two related fixes for the pgserve v2 instability:

- **#1575 (primary)** — pin `cwd` to genie's own package directory before the postgres.js pool's first connection so pgserve's `/proc`-walk fingerprinting always lands on `@automagik/genie` (one stable DB), not `app_anon_<fp>` ephemeral fallback or whichever foreign project the operator was `cd`'d into.
- **#1574 (defense in depth)** — extend the existing `cleanup` exit handler to fire-and-forget `sqlClient.end({timeout:1})` on hard exit, and add a `beforeExit` handler that awaits the full `shutdown()` for clean exits. Hook hot path was already daemon-routed in #1574's prior commit (63b2aa14); this closes the leak surface for the rest of the CLI subcommands.

## What changed

- `src/lib/db.ts` (+166/-2)
  - `resolveGeniePackageDir()` — walks up from `import.meta.dir` looking for `name === '@automagik/genie'`, depth-bounded, cached.
  - `pinCwdToGeniePackageDir()` exported helper for the daemon entrypoint; sets `daemonCwdPinned` so `_buildConnection` skips the cwd restore for the daemon's lifetime.
  - `_buildConnection` cwd-pin block: chdir to genie pkg dir → `pgModule({...})` → forced `await sqlClient`SELECT 1`` (so pgserve does its `/proc` walk under the pinned cwd) → `runPostConnectSetup` → finally restore cwd unless `daemonCwdPinned`.
  - Pool sizing gated on `skipBoot` (`GENIE_SKIP_DB_BOOT === '1'` || `isTestMode`): `max: skipBoot ? 1 : 50`, `idle_timeout: skipBoot ? 0 : 1`. Single persistent connection in CLI mode survives the cwd restore.
  - `cleanup()` extended with best-effort `sqlClient.end({timeout:1}).catch(()=>{})`.
  - `process.on('beforeExit', () => shutdown().catch(()=>{}))` registered for awaited drain on clean exit.

- `src/term-commands/serve.ts` (+28/-1)
  - Captures `operatorCwd` at start of `startForeground` (must read before the daemon's permanent pin).
  - Calls `pinCwdToGeniePackageDir()` BEFORE `claimServePidOrExit` / any DB call.
  - Hook socket `repoRoot` now reads `operatorCwd` so per-repo `.genie/hooks/` scans still find the operator's repo (not genie's package dir).

- `src/lib/db.test.ts` (+62)
  - Resolver round-trip test (verifies pin lands on a dir whose package.json declares `@automagik/genie`).
  - Source-shape asserts on `_buildConnection` strategy (pool sizing gate, SELECT 1 ordering before runPostConnectSetup, restore guard).
  - Source-shape asserts on cleanup pool drain + beforeExit wiring.

## Audit-log proof (the fix flipped identity)

Before:
\`\`\`json
{"event":"db_created","database":"app_anon_35da1b32060e","fingerprint":"35da1b32060e",
 "package_realpath":null,"name":null,"persist":false}
\`\`\`

After (this branch, `cd /tmp && dist/genie.js db query "SELECT 1"`):
\`\`\`json
{"event":"connection_routed","fingerprint":"a262844a108d","mode":"package",
 "package_realpath":"/home/genie/workspace/repos/genie/package.json",
 "name":"@automagik/genie"}
\`\`\`

Banner now reads `[pgserve] connected to app__automagik_genie_a262844a108d` (was `app_anon_<varies>`). All four required fields flip:
- `mode`: `script` → `package`
- `package_realpath`: null → genie's actual `package.json`
- `name`: null → `@automagik/genie`
- `persist`: false → true (DB is now GC-exempt)

## Test plan
- [x] `bun run typecheck` — clean
- [x] `bun run lint` — clean (pre-existing warnings only)
- [x] `bun test src/lib/db.test.ts` — 66 pass / 0 fail (3 new tests for the cwd-pin + exit-handler wiring)
- [x] `bun run build` — 5.57 MB bundle, 570 modules
- [x] Run `dist/genie.js ls` from `/tmp` (no package.json), verify banner shows `app__automagik_genie_<fp>` and audit log shows `mode:package` + `name:@automagik/genie` + `persist:true`
- [ ] Smoke test on a second machine after merge (this is the "two servers same version" case from the trace; the fix is identical-bundle-driven so should reproduce)
- [ ] Validate `genie serve` daemon path (manual: `genie serve start`, hit it from CLI subcommands, confirm all share the same pinned identity)

## Related
- Closes #1575 (database `app_anon_*` does not exist — one stable identity now)
- Amends #1574 (the `cleanup` extension + `beforeExit` close the non-hook leak surface; the hook hot path was already mitigated by 63b2aa14)
- Trace context: see issue comments on #1574 and #1575

### DIFF
--- a/src/lib/__tests__/edge-cases-stability.test.ts
+++ b/src/lib/__tests__/edge-cases-stability.test.ts
@@ -80,10 +80,15 @@ describe('7.1 Concurrent Operations', () => {
   test('connection pool config is sensible for concurrent operations', () => {
     const source = readFileSync(join(__dirname, '..', 'db.ts'), 'utf-8');
 
-    // max: 50 connections (serial CI; parallel test:parallel is a local-only tool)
-    expect(source).toContain('max: 50');
-    // Aggressive idle timeout (1s) to recycle unused connections
-    expect(source).toContain('idle_timeout: 1');
+    // max: 50 default for daemon/TUI/tests; short-lived CLI (GENIE_SKIP_DB_BOOT=1)
+    // drops to max:1 so the single fingerprinted connection survives the
+    // cwd restore (issue #1575). Operational justification (v4.260430.20):
+    // script-mode CLI fingerprints accumulated 296+ backends each,
+    // saturating pgserve max_connections=1000. Gate caps it at 1 per CLI.
+    expect(source).toContain('max: cliShortLived ? 1 : 50');
+    // Aggressive idle timeout (1s) for non-CLI; idle_timeout:0 keeps the single
+    // CLI connection alive for the process lifetime.
+    expect(source).toContain('idle_timeout: cliShortLived ? 0 : 1');
     // Socket mode gets the pgserve startup window; legacy TCP keeps the fast fallback.
     expect(source).toContain('connect_timeout: resolvePgConnectTimeoutSeconds(useSocket)');
     expect(source).toContain('if (!useSocket) return 5');
--- a/src/lib/db.test.ts
+++ b/src/lib/db.test.ts
@@ -1134,13 +1134,16 @@ describe('cwd pin for stable pgserve identity (issue #1575)', () => {
     }
   });
 
-  test('_buildConnection sources strategy uses skipBoot for max + idle_timeout', () => {
+  test('_buildConnection sources strategy uses cliShortLived for max + idle_timeout', () => {
     const source = readFileSync(join(__dirname, 'db.ts'), 'utf-8');
-    // Pool sizing is gated on skipBoot — short-lived CLI gets max:1 +
-    // idle_timeout:0 so the single fingerprinted connection persists
-    // across the cwd-restore in the finally block.
-    expect(source).toMatch(/max:\s*skipBoot\s*\?\s*1\s*:\s*50/);
-    expect(source).toMatch(/idle_timeout:\s*skipBoot\s*\?\s*0\s*:\s*1/);
+    // Pool sizing is gated on cliShortLived (NOT isTestMode — tests need
+    // concurrent connections; the pin/restore safety only matters when
+    // GENIE_SKIP_DB_BOOT=1 short-lived CLI subprocesses chdir back).
+    // Operational justification: v4.260430.20 saw script-mode CLI
+    // fingerprints accumulating 296+ pgserve backends each, saturating
+    // max_connections=1000. The gate caps that at 1 per subprocess.
+    expect(source).toMatch(/max:\s*cliShortLived\s*\?\s*1\s*:\s*50/);
+    expect(source).toMatch(/idle_timeout:\s*cliShortLived\s*\?\s*0\s*:\s*1/);
     // Forced SELECT 1 must run BEFORE runPostConnectSetup so pgserve
     // fingerprints under the pinned cwd.
     const selectIdx = source.indexOf('await sqlClient`SELECT 1`');
--- a/src/lib/db.test.ts
+++ b/src/lib/db.test.ts
@@ -1106,3 +1106,65 @@ describe('pgserve failure containment', () => {
     expect(selfHealCall).toBeLessThan(timeoutThrow);
   });
 });
+
+describe('cwd pin for stable pgserve identity (issue #1575)', () => {
+  test('pinCwdToGeniePackageDir resolves to genie package and chdirs', async () => {
+    const { pinCwdToGeniePackageDir } = await import('./db.js');
+    const beforePinCwd = process.cwd();
+    const result = pinCwdToGeniePackageDir();
+    try {
+      // pinned should point at a directory whose package.json declares
+      // name === '@automagik/genie'. If null, fall back gracefully — the
+      // resolver gave up but the call still returns the previous cwd.
+      expect(result.previous).toBe(beforePinCwd);
+      if (result.pinned !== null) {
+        const pkg = JSON.parse(readFileSync(join(result.pinned, 'package.json'), 'utf-8')) as {
+          name?: string;
+        };
+        expect(pkg.name).toBe('@automagik/genie');
+        expect(process.cwd()).toBe(result.pinned);
+      }
+    } finally {
+      // Restore for downstream tests in the same process.
+      try {
+        process.chdir(beforePinCwd);
+      } catch {
+        /* ignore */
+      }
+    }
+  });
+
+  test('_buildConnection sources strategy uses skipBoot for max + idle_timeout', () => {
+    const source = readFileSync(join(__dirname, 'db.ts'), 'utf-8');
+    // Pool sizing is gated on skipBoot — short-lived CLI gets max:1 +
+    // idle_timeout:0 so the single fingerprinted connection persists
+    // across the cwd-restore in the finally block.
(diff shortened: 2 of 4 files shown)


namastex888:
## Update — amendment commit `fcfb5c2f`

Just pushed an amendment that addresses two things together:

### 1. CI fix
The previous commit (`ff20f54b`) introduced `skipBoot = isTestMode || GENIE_SKIP_DB_BOOT === '1'` for pool-size gating. Two problems:
- **Variable collision**: `db.ts` already had a `skipBoot` constant for migration/seed/v1-prompt skipping (line 1476). The new declaration shadowed in scope but caused reader confusion.
- **CI deadlock**: including `isTestMode` in the gate meant `bun test` got `max:1`, deadlocking tests that need concurrent transactions. PG Tests shard 2/4 went red on `closeMergedTasks > moveTask` (15s timeout — the gate caused this).

Rename to `cliShortLived` and exclude `isTestMode`. Tests run against a dedicated test pgserve and skip the cwd restore anyway (`shouldRestoreCwd = !daemonCwdPinned && !isTestMode`), so they don't need the single-conn safety.

### 2. Why keep the gate at all (.20 operational evidence)
Got a fresh trace report from another user on `v4.260430.20`:
- `app_anon_f71d5a104ebe` (script-mode CLI fingerprint) accumulated **296 pgserve backends** from a small number of CLI processes.
- Total across all consumers reached `max_connections=1000` cap → `sorry, too many clients already` → cascade: `CONNECTION_DESTROYED`, circuit breakers, dropped events, mailbox retry failures.
- Even with the cwd-pin from this PR turning `app_anon_*` into `app__automagik_genie_<fp>`, a single CLI subprocess could still saturate if `max:50` per pool was honored. The gate caps that at 1 backend per short-lived CLI.

Dropping the gate entirely (which the user briefly considered after I tried updating the source-shape test) would re-open the saturation surface for any non-hook CLI invocation that escapes the daemon-UDS routing in PR #1574.

### 3. Test updates
- `src/lib/db.test.ts:1137-1143` — source-shape assertion now matches `cliShortLived` form.
- `src/lib/__tests__/edge-cases-stability.test.ts:80-90` — assertion updated from literal `max: 50` to ternary `cliShortLived ? 1 : 50`. The gate is justified by operational evidence; the test should reflect that.

### 4. Known follow-ups (NOT in this PR — separate fixes)
A `/trace` on the same `.20` report uncovered a related but independent bug:

**`null is not a function` in scheduler reconciler** (`src/lib/scheduler-daemon.ts:1721, 1824`):
- Root cause: `src/lib/db.ts:1695` `_buildConnection` returns the module-level `sqlClient` rather than a local. Under concurrent rebuild, `healthCheckCachedClient` can null the global between assign and return → caller awaits `null` as tagged-template fn → exact error string.
- Fix: bind to local first in `_buildConnection`, return the local. ~10 lines.
- Confidence: high.
- Recommend: separate PR (or fold into this one in a third commit if reviewers prefer).

**Filewatch FK violation** (reporter claimed `session_content_session_id_fkey`):
- Trace finding (medium-low confidence): the in-source ingest path correctly sequences parent before child. The reported FK constraint name appears to be a generic log line that doesn't actually surface the real constraint. Recommend a 1-line diagnostic patch to log `err.constraint_name`, then re-trace once we have a confirmed failure.

### Verified
- `bun test src/lib/db.test.ts src/lib/__tests__/edge-cases-stability.test.ts src/lib/task-close-merged.test.ts`: 134/134 pass
- `bun run typecheck`: clean
- `bun run build`: 5.6 MB

Pre-existing 5s-timeout `(unnamed)` flakes in the full suite are ordering-dependent and unrelated to the pool gate (they pass in isolation, fail in full-suite under specific test orders).

PR #1684, opened 2026-05-07
fix(registry): preserve native_team_enabled + provider across ON CONFLICT in register()

### LINKED ISSUE #1682: agent-registry register() ON CONFLICT silently drops native_team_enabled + provider — every claude worker misrouted to tmux send-keys
## Summary

`register()` at `src/lib/agent-registry.ts:293` has an ON CONFLICT DO UPDATE SET clause that silently drops `native_team_enabled`, `native_agent_id`, `native_color`, `parent_session_id`, `provider`, and `transport`. Because `findOrCreateAgent` (`agent-registry.ts:1030`) writes the identity row *first* with column defaults (`native_team_enabled=FALSE`, `provider=NULL`), the subsequent `register(workerEntry)` call is a **no-op for those fields** — the spawn-intent values never reach the row.

Result: every claude-native worker dispatched after commit `dfc875ef` ("feat(spawn): collapse to single UUID row — wish G3", Mon 2026-05-04, part of the in-flight retire-session-names-id-only wish) is stamped non-native at the DB level. `protocol-router.ts:405` reads `worker.nativeTeamEnabled` → `false` → falls through to `injectToTmuxPane()`, which types the message into the agent's TUI input via `tmux send-keys` (`protocol-router.ts:651-655`).

This is the **same anti-pattern** as the omni-bridge double-delivery race in #1679 — body → 200 ms settle → Enter via send-keys, vulnerable to TUI bootstrap timing. `claude-code.ts:475-477` even has a comment cross-referencing the duplication: *"Two-phase send-keys with a 200 ms settle… Matches injectToTmuxPane in protocol-router.ts."*

## Root cause

`src/lib/agent-registry.ts:293` ON CONFLICT clause:

```sql
ON CONFLICT (id) DO UPDATE SET
  pane_id = EXCLUDED.pane_id,
  session = EXCLUDED.session,
  state = EXCLUDED.state,
  last_state_change = EXCLUDED.last_state_change,
  team = COALESCE(agents.team, EXCLUDED.team),
  role = COALESCE(agents.role, EXCLUDED.role),
  custom_name = COALESCE(agents.custom_name, EXCLUDED.custom_name),
  updated_at = now()
```

Missing: `provider`, `transport`, `native_agent_id`, `native_color`, `native_team_enabled`, `parent_session_id`.

## Evidence — live DB probe

Three current `team=genie` claude workers in this workspace, all misrouted:

| name | role | team | native_team_enabled | provider |
|------|------|------|---------------------|----------|
| `fix` | fix | genie | **false** | **null** |
| `fix-bug2` | fix-bug2 | genie | **false** | **null** |
| `engineer` | engineer | genie | **false** | **null** |

The `agent_templates` table has the correct values for all four (`native_team_enabled=true`, `provider='claude'`). The spawn intent is correct; the agents-table write is corrupt.

## Causal chain

1. `genie agent spawn fix --team genie` runs (`agents.ts:2640+`)
2. `resolveNativeTeam` builds `nativeTeam.enabled=true` for claude (`agents.ts:1844`)
3. `findOrCreateAgent('fix','genie','fix')` INSERTs the identity row with column defaults — `native_team_enabled=false`, `provider=null`
4. `registerSpawnWorker` builds `workerEntry { nativeTeamEnabled: true, provider: 'claude', native_agent_id: 'fix@genie' }` (`agents.ts:662, 673`)
5. `registry.register(workerEntry)` hits the buggy ON CONFLICT — UPDATE list omits native_* + provider + transport. Row keeps step-3 defaults.
6. Orchestrator calls `genie agent send 'msg' --to fix`
7. `protocol-router.deliverToWorker` (line 394) reads `worker.nativeTeamEnabled` → false → `injectToTmuxPane` (line 408)
8. `injectToTmuxPane` runs `tmux send-keys -t %<pane> '<body>'`, sleeps 200 ms, then `send-keys Enter` — body typed into Claude Code's TUI input
9. Native-inbox fallback at `protocol-router.ts:411-420` never fires — `send-keys` returns success on a living pane regardless of inner program state

### Primary (one SQL change at `agent-registry.ts:293`)

Extend the ON CONFLICT DO UPDATE SET clause:

```sql
native_team_enabled = EXCLUDED.native_team_enabled,
native_agent_id     = COALESCE(EXCLUDED.native_agent_id, agents.native_agent_id),
native_color        = COALESCE(EXCLUDED.native_color, agents.native_color),
parent_session_id   = COALESCE(EXCLUDED.parent_session_id, agents.parent_session_id),
provider            = COALESCE(EXCLUDED.provider, agents.provider),
transport           = COALESCE(EXCLUDED.transport, agents.transport)
```

Why `EXCLUDED` (not COALESCE) for `native_team_enabled`: a respawn under different protocol settings must be allowed to re-flag the row. Spawn intent is authoritative.

Why `COALESCE` for the others: preserves immutable identity values when later callers omit them.

### Backfill (one-time, post-merge)

```sql
UPDATE agents a
SET native_team_enabled = true,
    provider = COALESCE(a.provider, t.provider)
FROM agent_templates t
WHERE a.custom_name = t.name
  AND a.team = t.team
  AND t.native_team_enabled = true
  AND a.state IS NOT NULL;
```

### Test gap

`src/lib/protocol-router.test.ts:93` hardcodes `nativeTeamEnabled=false` in a fresh INSERT — never exercises the ON CONFLICT preservation path. Add a test that calls `findOrCreateAgent` first, then `register` with `nativeTeamEnabled=true`, then asserts the row read back has the field true.

## Companion (separate, hardening — not required to ship the SQL fix)

Once the registry fix lands and is backfilled, `injectToTmuxPane` (`protocol-router.ts:639`, module-private) should never be reached by claude workers. Either:
- Gate it: `if (worker.provider !== 'codex') return false;` — forces native-inbox fallback for claude.
- Or remove it for claude entirely, since the fallback path at `protocol-router.ts:411-420` already handles delivery.

This also retires the documented body→200ms→Enter race surface — the same anti-pattern as #1679 Bug 1.

## Affected scope

- **Every claude-native agent registered after `dfc875ef` (Mon 2026-05-04)** — i.e., the entire retire-session-names-id-only wish family (PG tasks #175/#176/#177).
- Confirmed misrouted: `engineer`, `fix`, `fix-bug2` in `team=genie` (this workspace).
- Impacts: `genie send`, `genie agent send`, team-lead → teammate dispatches, resume-context injection at `protocol-router-spawn.ts:416`, team-leader auto-spawn at line 268.
- Codex agents unaffected by the misroute (correctly send-keys-targeted today).
- **Recommend folding the SQL fix into wish G2 (`#177`)** since the same wish introduced the regression.

## Confidence

**HIGH.**
- DB rows directly show the misroute precondition.
- Code shows the SQL UPDATE clause omits exactly the relevant columns (verified in this issue's source link above).
- Routing branch at `protocol-router.ts:405` is the only consumer of `worker.nativeTeamEnabled` and matches symptom one-to-one.
- Function-name verification: `register` exported at `agent-registry.ts:255`, `findOrCreateAgent` at `:1030`, `sendMessage` at `protocol-router.ts:555`. `injectToTmuxPane`, `writeToNativeInbox`, `deliverToWorker` are module-private.

🤖 Investigation by `trace-misroute@genie`, synthesized + filed by `genie-omni`. Related: #1679.

namastex888:
## Summary

Closes #1682.

`register()` in `src/lib/agent-registry.ts` was no-op'ing the protocol-related fields on its `ON CONFLICT (id) DO UPDATE` path. After `findOrCreateAgent` created the identity row with default values (`native_team_enabled=false`, `provider=NULL`, etc.), the subsequent `register()` carrying the real spawn intent left those defaults pinned — and routing kept reading `nativeTeamEnabled === false`, dispatching claude workers down `injectToTmuxPane` (the body→200ms→Enter race surface) instead of `writeToNativeInbox`.

Three live workers observed in this misrouted state: `engineer`, `fix`, `fix-bug2`.

## Causal chain

1. `findOrCreateAgent('engineer', 'genie', 'engineer')` — INSERT with `native_team_enabled DEFAULT false`, `provider IS NULL`.
2. Spawn pipeline calls `register({ ...id, nativeTeamEnabled: true, provider: 'claude', transport: 'tmux', nativeAgentId: 'engineer@genie', ... })`.
3. INSERT collides on `(id)`; `DO UPDATE SET …` refreshes only `pane_id, session, state, last_state_change, team, role, custom_name, updated_at` — none of the protocol fields.
4. `agents.native_team_enabled` stays `false`, `agents.provider` stays `NULL`.
5. Routing in `protocol-router` reads stale row → `injectToTmuxPane` for claude workers.

## Fix

Extend the `ON CONFLICT … DO UPDATE SET` clause at `src/lib/agent-registry.ts:293`:

- `native_team_enabled = EXCLUDED.native_team_enabled` — spawn intent is authoritative; a respawn under different protocol settings must be allowed to re-flag the row.
- `native_agent_id`, `native_color`, `parent_session_id`, `provider`, `transport` use `COALESCE(EXCLUDED.<col>, agents.<col>)` — preserve immutable identity values when later callers (heartbeat-style refreshes) omit them.

Regressing commit: `dfc875ef`.

## Tests

Two new regressions added in `src/lib/agent-registry.test.ts` under a fresh top-level (non-skipped) `describe` block:

1. `register preserves nativeTeamEnabled + provider across findOrCreateAgent → register` — drives the exact misroute path. Asserts `native_team_enabled` flips `false → true` (EXCLUDED) and `provider/transport/native_agent_id/native_color/parent_session_id` populate from NULL (COALESCE).
2. `register preserves established protocol fields when later caller omits them` — confirms the COALESCE branches don't clobber values when a later refresh leaves the fields undefined.

These use real UUIDs (via `randomUUID()` from `findOrCreateAgent`) so migration 061's `agents_id_shape_check` accepts the inserts. The existing legacy `describe.skip` block above is untouched (still pending wish #175 fixture rewrite).

## Backfill SQL — operator runs post-merge (NOT in this PR)

```sql
UPDATE agents a
SET native_team_enabled = true,
    provider = COALESCE(a.provider, t.provider)
FROM agent_templates t
WHERE a.custom_name = t.name
  AND a.team = t.team
  AND t.native_team_enabled = true
  AND a.state IS NOT NULL;
```

Idempotent — restores correctness for currently-misrouted live workers.

## Coordination

Courtesy ping for wish #175 G2 #177 (`retire-session-names-id-only`). Same file, different region — G2 touches `reconcileStaleSpawns` (line 604+); this fix is at `register()` line 293. No conflict observed against `origin/wish/retire-session-names-id-only` at fetch time. Happy to fold into G2 if cleaner.

## Validation

- `bun run typecheck` → 0 errors ✅
- `bunx biome check src/lib/agent-registry.ts src/lib/agent-registry.test.ts` → clean ✅

**Test-gate caveat:** `bun test src/lib/agent-registry.test.ts` hung in this workspace (unrelated environmental issue — collateral from in-flight dogfood work, see #1682 thread). Reviewer should run the test pre-merge to confirm the two new regression tests pass.

## Out of scope

- Companion `injectToTmuxPane` retirement (tracked separately as a hardening ticket).
- Any rework of `findOrCreateAgent` (works correctly; bug is purely on the register path).
- Codex worker behavior (unaffected).

## Test plan

- [ ] Reviewer: `bun test src/lib/agent-registry.test.ts -t "1682"` — both new tests pass.
- [ ] Reviewer: confirm full `bun test src/lib/agent-registry.test.ts` stays green (no regression in the existing `describe.skip` block which remains skipped pending #175).
- [ ] Operator: run backfill SQL above against any environment with mid-flight misrouted workers.


namastex888:
Addressed @gemini-code-assist and @chatgpt-codex-connector review in `b784aa27`:

- **Transport COALESCE no-op (both bots, correct)**: Changed VALUES to `${agent.transport ?? null}` and DO UPDATE to `COALESCE(EXCLUDED.transport, agents.transport, 'tmux')`. Fresh INSERT with no transport still defaults to `'tmux'`; a heartbeat refresh that omits the field now preserves the established value (e.g. `'inline'`).
- **`native_team_enabled` (Gemini)**: Author's design intent stands — spawn intent is authoritative on this field. Tightened the ON CONFLICT comment to make the caller contract explicit: every `register()` call MUST set `nativeTeamEnabled` because the `?? false` fallback in VALUES will clobber an existing `true` if omitted. Surveyed the three current runtime callers (`registerSpawnWorker`, `session.ts`, `protocol-router-spawn.ts`) — all set the field explicitly.
- **Test coverage gap (both bots, implicit)**: Added a third regression test (`register preserves non-default transport across heartbeat refresh (PR #1684 review)`) that registers with `transport: 'inline'`, then heartbeat-refreshes omitting the field, asserting it stays `'inline'`. Pre-fix this would have silently flipped to `'tmux'`.

Validation: `bun run typecheck` clean, `bunx biome check` clean. `bun test` hangs in this workspace (same environmental issue documented in PR body).

PR #2432, opened 2026-05-16
feat(db): genie-dedicated-role-cutover Wave 3 — default-on + doctor + matrix (Goal A G4)

namastex888:
## Wave 3 — Group 4 (final): flip the cutover LIVE

Makes the dedicated-role cutover the **default** behind a documented kill-switch. genie stops connecting as the `postgres` superuser on the load-bearing direct-postmaster path.

### Honest framing (council-mandated)
Least-privilege **integrity containment** — *a genie failure can no longer DROP omni, exhaust the shared cluster, or corrupt shared WAL*. **NOT confidentiality**: `--auth-local=trust` and the passwordless `postgres` superuser are unchanged (producer-side, out of scope). Zero bytes move.

### Deliverables
- **Default-on gate, single source of truth.** `isRoleCutoverEnabled()` in `role-cutover.ts`: `GENIE_ROLE_CUTOVER=0` (and only `0`) is the documented kill-switch → byte-identical legacy `postgres`/`postgres`; unset/`1`/other ⇒ ON. Both `db.ts` gate sites consume it so the literal cannot drift.
- **`genie doctor --connection-identity`** (read-only): role, `rolsuper` (false when active), database, GRANT summary, fallback flag+reason, per-fingerprint sentinel. SELECT-only, zero mutations. `--json` supported.
- **`role-cutover.matrix.test.ts`**: N=8 concurrent single-provision (cold + warm storm), fallback trio, multi-fingerprint non-stranding, test-mode/FORCE_TCP/accept-hook skip, kill-switch + default-on contract.
- **`docs/_internal/role-cutover-note.md`**: honest framing, kill-switch, fallback-is-rollback, doctor reading guide. *(Content authored on disk; cross-repo `.docs-vendor` submodule PR + pointer bump is the documented human-gated docs workflow — see Concerns.)*

### Wave 2 safety preserved
Direct-postmaster rebind only (no accept-hook/router/TCP/test widening); non-hard-fail fallback; per-fingerprint sentinel; test-mode + `GENIE_PG_FORCE_TCP` SKIP. In test mode `shouldAttemptRoleCutover` returns false → the flip adds zero connections under the suite.

### Validation
- `bun run typecheck` clean · `bun run dead-code` (knip) clean · `bun run lint` exit 0
- Targeted set (matrix + db.role-cutover + provision + migration-replay + db.test): **106 pass / 0 fail**
- **Full-suite regression check:** pristine base = **14** failures; with Wave 3 = **14** failures — identical ~5000ms pgserve-saturation timeouts across unrelated DB-backed suites. **No regression** from the flip.
- db.ts source-guard green (only the gate predicate changed; no connection literals / `buildPgClientOptions` signature touched).

### DIFF
--- a/src/genie-commands/doctor.ts
+++ b/src/genie-commands/doctor.ts
@@ -22,10 +22,11 @@ import { fileURLToPath } from 'node:url';
 import { $ } from 'bun';
 import type { BridgeStatusResult, PingOptions } from '../lib/bridge-status.js';
 import { contractClaudePath, getClaudeSettingsPath } from '../lib/claude-settings.js';
-import { isAvailable as isRuntimePgAvailable } from '../lib/db.js';
+import { getConnection, isAvailable as isRuntimePgAvailable, resolveDatabaseName } from '../lib/db.js';
 import { tmuxBin } from '../lib/ensure-tmux.js';
 import { genieConfigExists, getGenieConfigPath, isSetupComplete, loadGenieConfig } from '../lib/genie-config.js';
 import { respawnInvocation } from '../lib/respawn.js';
+import { type RoleCutoverInspection, inspectRoleCutover } from '../lib/role-cutover.js';
 import { checkCommand } from '../lib/system-detect.js';
 import { findWorkspace } from '../lib/workspace.js';
 import {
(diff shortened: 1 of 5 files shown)


namastex888:
Duplicate of #2433 (same branch feat/role-cutover-w3-default-on). Consolidating on #2433.

PR #2485, opened 2026-05-24
fix(migrations): 002 discovers canonical pgserve port instead of hardcoding 8432

automagik-genie:
## Problem
Migration `002-kill-embedded-pgserve-legacy` hardcoded `CANONICAL_PORT = 8432` and stops any postgres listening on a *different* port. After the autopg-v3 socket-singleton cutover the canonical backbone binds **5432**, which makes this migration:

1. **Inert** on a healthy 5432 host (it probes 8432 for "canonical", finds nothing, no-ops); and worse
2. **Actively dangerous** in the mixed window where a stray legacy postmaster is still live on 8432 alongside canonical 5432 — it treats 8432 as canonical and **stops the real 5432 backbone** as "legacy".

This affects any user upgrading across the 8432→5432 cutover, not just one host.

## Fix
Discover the canonical port from the autopg/pgserve binary at runtime (`<bin> status --json` → `.port`), never hardcode it. When no canonical binary is installed or it can't report a port, the migration is a **safe no-op** — we never stop a postmaster we can't positively distinguish from canonical.

- Extracted `selectLegacyEmbedded(canonicalPort, listening)` as a pure, unit-tested helper.
- Memoized port resolution for the migration run.

## Tests
`test/migrations/002-kill-embedded-pgserve-legacy.test.ts` (5 cases): never selects canonical on a 5432-only host; selects the stray 8432 (not 5432) in the mixed window; no-op when canonical port is unidentifiable (null); legacy-era 8432 no-op.

### DIFF
--- a/src/genie-commands/update.ts
+++ b/src/genie-commands/update.ts
@@ -20,6 +20,7 @@ import {
 import { homedir } from 'node:os';
 import { dirname, join } from 'node:path';
 import { genieConfigExists, loadGenieConfig, saveGenieConfig } from '../lib/genie-config.js';
+import { extractPgservePortFromStatus } from '../lib/pgserve-status.js';
 import { VERSION } from '../lib/version.js';
 import { pm2ProcessNameCandidates, regenerateEcosystemConfig } from './install.js';
 import { type CleanupReport, cleanupLegacyArtifacts, parseSkipCleanupFlag } from './legacy-cleanup.js';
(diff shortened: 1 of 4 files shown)


automagik-genie:
Thanks for the reviews — all three points addressed in `9b0a1a0`:

- **Codex (P1, nested port shapes):** `resolveCanonicalPort` now reuses `extractPgservePortFromStatus` (moved to `src/lib/pgserve-status.ts`, re-exported from `update.ts`), so it tolerates top-level `port` **and** nested `instance.port` / `runtime.port`. No more permanent no-op on hosts emitting the nested shape.
- **Gemini (port parsing):** added a `port > 0` guard so `Number(null|false|'')===0` is rejected.
- **Gemini (stop all):** `selectLegacyEmbedded` now returns **every** non-canonical postmaster (filter, not find); `apply` stops each and `validate` waits for all to exit. New test covers the multi-stray case.

PR #2502, opened 2026-07-02
fix!: re-arm hook dispatch (in-process default) — branch-guard was silently open

namastex888:
## Summary

Fixes a HIGH defect discovered while building omni approvals: **v5's hook dispatch has been silently falling open since the demolition merged**, so branch-guard — the standing §19 merge law — and all hooks have not been running on the real path. Executes wish [`dispatch-inproc-default`](.genie/wishes/dispatch-inproc-default/WISH.md).

## The bug

The demolition deleted the hook daemon (`src/serve/`), but `dispatch-command.ts` still defaulted to a socket client (`dispatch-client.ts`) that connects to that now-absent daemon. On the socket miss it "fails open" — empty stdout = allow-by-default — running **no handlers**. Proven: `git push origin main` was denied only under `GENIE_HOOK_FORCE_INPROC=1` and sailed through empty on the default path; `~/.genie/hook-fallback.log` logged `connect error: ENOENT` on every hook. The entire daemon indirection existed for a Postgres connection leak (#1574) that died with Postgres — orphaned harness the demolition missed.

## The fix

- **In-process is now the unconditional default** (the path that provably runs the real registry). Deleted the orphaned `dispatch-client.ts` + its dead `redaction.ts` dependency + the `FORCE_INPROC`/`SKIP_DB_BOOT` fossils + the daemon build target. Net −536 lines.
- **Fail CLOSED at the entry level** (the actual leak): bad stdin and uncaught throws now emit a non-empty non-allow envelope instead of empty-allow — while preserving the `AskUserQuestion` inline-picker carve-out (an unknown-tool error must not emit a `hookSpecificOutput` envelope, or it'd suppress the picker).
- **Permanent regression gate**: a subprocess test drives the *built* `genie hook dispatch` on the default path and asserts `git push origin main` is DENIED. The reviewer proved it fails hard on a simulated fall-open — closing the blind spot that let this hide (every prior test drove `dispatch()` directly, never the entry CC invokes).

## Proof

`git push origin main` → `permissionDecision:"deny"` on the built dist's default path (was silently allowed). branch-guard is re-armed and regression-gated. 488 `bun run check` tests, e2e, knip all green.

## Coordination

`wish/omni-runner-port` (separate branch) also edits `src/hooks/index.ts` in the `buildDenyResponse`/line-225 region — a deliberate 3-way merge is needed when both land (keep this branch's `buildFailClosedResponse` AND omni's response-builder changes).

## Discovered follow-up (not in this PR)

CLAUDE.md broadly documents the deleted v4 architecture (agent/team/exec namespaces, worker registry, OTel, "~305KB bundle" vs the real 0.92MB) — a `CLAUDE.md-for-v5` rewrite is warranted separately.

Merge decision: Felipe.


namastex888:
## Review response

**Gemini (HIGH — `beforeAll` pipe-buffer deadlock): accepted & fixed** in `19ce1d05`. Switched the regression-gate build to `Bun.spawnSync` (buffers output automatically), removing the deadlock risk when `bun run build` writes past the ~64KB pipe buffer.

**Codex (P1 — plugin doesn't register `genie hook dispatch` for Bash): factually correct, but the proposed fix is declined** — with reasoning:

- The observation is right: `plugins/genie/hooks/hooks.json` wires dispatch to PreToolUse `SendMessage` only, not `Bash`.
- But adding a Bash registration to the plugin would impose genie's **internal §19 merge law** (`no push to main`) on *every end-user repo that installs the genie plugin* — that is genie's own development policy, not something to force on arbitrary user repos.
- The place branch-guard must fire is the **genie repo itself**, where `.claude/settings.json` routes all tools to dispatch via its `*` PreToolUse matcher (Bash → `genie hook dispatch` → branch-guard). Verified: `git push origin main` is denied on that path. That is the exact scenario this wish fixed (agents working in the genie repo were bypassing branch-guard because the dispatcher fell open).
- The regression test correctly validates the **dispatcher's decision**, which is this wish's subject — not per-host registration wiring. Added a docstring clarifying that scope so the test can't be read as claiming plugin-wide coverage.

Whether the genie plugin should route additional events to dispatch for end users is a separate registration/design question, not a defect in this fix.

PR #2504, opened 2026-07-02
feat: v5 completion — CLAUDE.md truth, Codex launch target, 5.x distribution

namastex888:
## Summary

The final umbrella group of the v5 lightweight body — three independent tracks, landed as three separate commits. Executes wish [`v5-completion`](.genie/wishes/v5-completion/WISH.md), each group independently SHIP-reviewed.

## Three commits

**`da5185ba` — docs(v5): CLAUDE.md rewrite + drift guard**
The instruction set every agent loads now describes v5, not the deleted v4 harness. Removed the `agent/team/exec` namespaces, pgserve, `GENIE_OTEL_PORT`, native-teams, tmux, "~305KB"; documented the real 12 commands, per-repo + global `genie.db`, in-process fail-closed hook dispatch, 4 runtime deps (~1.3MB bundle). A drift-guard test (mutation-proven to bite) locks out 12 fossil tokens. OTel *relay* (Codex, `codex-config.ts:14318`) is documented, not purged.

**`cd3c126d` — feat(v5): `genie launch --agent codex` + Hermes decision**
A single `AGENT_COMMANDS` mapper parameterizes the launch pane command: `claude` byte-identical, `codex` → `codex exec "$(cat …)"` (verified non-interactive — emits-and-runs, won't hang). Invalid `--agent` → typed error before any side effect. Hermes decided in a `.genie` doc (defer + future omni-shaped runner; an API agent doesn't fit the terminal-pane model) — no half-built launcher.

**`e92ad2f1` — chore(v5)!: 5.x version scheme + remove npm vestiges**
Corrects a false premise the plan review caught: npm was discontinued 2026-05-09 — v5 ships cosign/SLSA signed tarballs via curl. So: 5.x scheme (`5.YYMMDD.N`) across `scripts/version.ts` + the load-bearing `version.yml` CI derive step; `genie --version` → `5.260702.1`; **removed** the npm vestiges (`publishConfig`, dead `prepack`) per the housekeeping handoff; release workflows audited (fixes + deferrals in release-checklist.md); release-bot and branch-guard confirmed to be different planes (dev-only bot push; the local hook never loads on CI).

## Gates

Full `bun run check` (576 tests) + build + e2e green; dead-code clean; the three tracks touch disjoint files.

## Notes

- Mixed-concerns by design (user requested one wish) — landed as 3 reviewable/revertible commits.
- Follow-up (out of scope, noted): a codex `--sandbox` passthrough so real panes get write access (bare `codex exec` respects user config, defaults may be read-only); plugin-version JSONs resync to 5.x on the next CI release.

This is the last v5 umbrella group — its merge completes the lightweight body end to end.

Merge decision: Felipe.

### DIFF
--- a/.genie/wishes/v5-completion/WISH.md
+++ b/.genie/wishes/v5-completion/WISH.md
@@ -66,9 +66,9 @@ Close out the v5 lightweight body with the three remaining independent tracks, b
 2. Drift-guard: a test or e2e grep gate asserting `CLAUDE.md` contains none of the retired-fossil tokens.
 
 **Acceptance Criteria:**
-- [ ] Zero stale-v4 fossils (grep gate incl. the native-teams tokens); real 12-command surface + genie.db + in-process hooks documented; still-true sections preserved; OTel relay reference (if any) untouched.
-- [ ] Drift-guard gate fails hard if a fossil returns.
-- [ ] `bun run check` green.
+- [x] Zero stale-v4 fossils (grep gate incl. the native-teams tokens); real 12-command surface + genie.db + in-process hooks documented; still-true sections preserved; OTel relay reference (if any) untouched.
+- [x] Drift-guard gate fails hard if a fossil returns.
+- [x] `bun run check` green.
 
 **Validation:**
 ```bash
--- a/.genie/wishes/v5-completion/WISH.md
+++ b/.genie/wishes/v5-completion/WISH.md
@@ -2,7 +2,7 @@
 
 | Field | Value |
 |-------|-------|
-| **Status** | DRAFT |
+| **Status** | DONE — all 3 groups SHIP-reviewed (2026-07-02); 3 commits |
 | **Slug** | `v5-completion` |
 | **Date** | 2026-07-02 |
 | **Author** | Felipe + Genie |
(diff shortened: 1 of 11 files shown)


namastex888:
## Review response

**Gemini MEDIUM (in-operator agent lookup): accepted** — `resolveAgentCommand` now uses `name in AGENT_COMMANDS` + a `keyof typeof` lookup instead of the `Record` cast. Type-safe, no assertion.

**Gemini MEDIUM (drift gate missing native-teams/mailbox): accepted** — added both to `RETIRED_FOSSILS` (verified neither appears legitimately in the rewritten CLAUDE.md, so no false-trip).

**Gemini HIGH (removing the stamp step breaks the compiled binary's version): declined — verified false.** The concern assumes the compiled binary reads `package.json` at runtime, but `src/lib/version.ts:65` has a dedicated `bun --compile` path that reads a `VERSION` stamp file dropped beside the binary *first* (because `import.meta.dir` is `bunfs` in a compiled binary), and `scripts/build-binary.sh` writes that `VERSION` file into the release tarball (stamped from `package.json`). Proven empirically — I built the real artifact and ran it:
```
$ bash scripts/build-binary.sh --platform darwin-arm64
$ tar xzf dist/genie-5.260702.1-darwin-arm64.tar.gz && ./genie --version
5.260702.1
```
The compiled binary reports the correct 5.x version and never touches `package.json`. The stamp step G3 removed was a genuine no-op — it rewrote an `export const VERSION = '…'` literal that `src/lib/version.ts` no longer contains (it resolves at runtime).

**Codex P2 (npm pack no longer builds dist without prepack): declined — npm is not a v5 distribution path.** npm publishing was discontinued 2026-05-09; v5 ships cosign/SLSA-signed tarballs via `curl … install.sh`, and the release artifact is built by `scripts/build-binary.sh` (`bun build --compile`), which builds its own binary (proven above) — it does not use `npm pack`/`prepack`. Re-adding `prepack` "so `npm pack` is self-contained" would restore the exact npm vestige this group was tasked (by the v5-housekeeping handoff) with removing. The `npm pack --dry-run` in the release checklist runs after `bun run build` purely to validate the `files` allowlist, not as a shipping path.

Fixes in 869dc229.

PR #2514, opened 2026-07-04
feat(omni): Model A — persona + session + streaming + ⏳→✅ ack for genie omni serve

namastex888:
## What

Enhances `genie omni serve` (the NATS bridge letting an Omni WhatsApp instance talk to a Claude Code agent) from a bare `claude -p "<msg>"` into **Model A**: persona injection, session continuity, streaming, and a ⏳→✅/❌ reaction ack on the user's message.

## Changes

**`src/lib/omni-runner.ts`**
- `defaultSpawnClaude` now spawns `claude -p --output-format stream-json --verbose --session-id <uuid> [--append-system-prompt-file <persona>] <message>` and parses the final reply out of the NDJSON stream.
- New pure, exported helpers (unit-tested without forking):
  - `buildClaudeArgs({message, sessionId, personaFile?})` — the argv contract.
  - `extractStreamJsonReply(raw)` — reply extraction: prefer the terminal `{"type":"result","subtype":"success","result"}` event → fall back to concatenated `assistant` text blocks → fall back to raw stdout on non-stream-json output.
  - `deterministicSessionId(instance, chat)` — stable v5-shaped UUID (sha256 of the pair) so a conversation resumes across messages, with no state to persist.
- `SpawnClaudeOpts` gains optional `personaFile` + `sessionId`. `runOneShot` resolves `personaFile = route.persona ?? <repo>/AGENTS.md (if it exists)` and passes the deterministic session id.
- The inbound WhatsApp stanza id (`messageId`) is threaded `handleMessage → startRoutedRun → runOneShot`, and a route-scoped ⏳→✅/❌ reaction is set on it: ⏳ right before the spawn, ✅ once a genuine reply is published, ❌ on non-zero exit / timeout / crash.
- The existing approval-scoped `emitStatusReaction` is generalized into a shared `emitReaction` seam; the route ack reuses it but records **no** approval glyph and skips the reconciliation guard (the ⏳→✅/❌ pair fires exactly once per run). It stays fire-and-forget (drained by `whenIdle`) and never throws.

**Config** — `OmniRoute` gains optional `persona` (absolute path) in both the Zod schema (`src/types/genie-config.ts`) and the runtime type (`src/lib/omni-config.ts`).

## How the ack is wired (route-scoped + non-blocking)

The ack targets `{instance: route.instance, chat: route.chat, messageId: <inbound stanza id>}` — the route's own chat, **not** `config.approvalChat`. It goes through the same injectable `setReaction` seam as the approval flow, is added to `inFlightReactions` and drained by `whenIdle`, and the run/publish never awaits the HTTP react. A missing `messageId` is a no-op (no reaction at all).

## Session-id derivation

`sha256("omni-session:<instance>:<chat>")`, first 128 bits, shaped into a valid v5-style UUID (version nibble `5`, variant `8/9/a/b`). Same `(instance, chat)` ⇒ same id ⇒ conversation continuity; deterministic so two hosts serving the same route converge.

## Constraints honoured

- **Approval flow untouched** — same ⏳/✅/❌ behaviour, glyph recording, and reconciliation. `emitStatusReaction` keeps its exact prior semantics (glyph-recorded + guarded) as a thin wrapper over the shared seam.
- Omni-side not touched.

## Tests

12 new tests (injected `spawnClaude` + `setReaction`, zero fork/HTTP): argv contract; stream-json parsing (result / assistant-delta / raw fallback); session-id stability + shape; route-scoped ⏳→✅/❌ acks; ❌ on non-zero exit and on timeout; no reaction when `messageId` absent; a failed reaction never blocks the publish; persona resolution (explicit / AGENTS.md fallback / none).

`bun run check` (typecheck + biome + knip + tests): **green** — 647 pass / 1 skip / 0 fail.

## Summary by CodeRabbit

* **New Features**
  * Added optional route-specific personas: when configured, the runner appends the persona/AGENTS content to the agent prompt, falling back to a repo-level persona file when present (invalid persona paths are ignored).
  * Improved continuity by using deterministic session context per route and chat for one-shot runs.
  * Enhanced routed-run status updates with a clear ⏳ → ✅/❌ lifecycle correlated to the originating inbound message.
* **Bug Fixes**
  * Improved reliability of final replies when Claude outputs streamed `stream-json`/NDJSON, with safe passthrough fallback.
  * Implemented resume-first session handling to reduce session-related failures; improved handling of “soft” stream failures.

### DIFF
--- a/src/lib/omni-config.ts
+++ b/src/lib/omni-config.ts
@@ -38,6 +38,12 @@ export interface OmniRoute {
   instance: string;
   chat: string;
   repo: string;
+  /**
+   * Absolute path to a persona / AGENTS.md file appended to the agent's system
+   * prompt for this route's runs. Omitted → the runner falls back to
+   * `<repo>/AGENTS.md` if it exists, else no persona.
+   */
+  persona?: string;
 }
 
 /**
--- a/src/lib/omni-runner.test.ts
+++ b/src/lib/omni-runner.test.ts
@@ -15,12 +15,14 @@ import { join } from 'node:path';
 import type { OmniRuntimeConfig } from './omni-config.js';
 import {
   type OmniSend,
+  type RawClaudeSpawn,
   type SpawnClaude,
   type SpawnClaudeResult,
   buildClaudeArgs,
   createOmniRunner,
   deterministicSessionId,
   extractStreamJsonReply,
+  runClaudeSession,
 } from './omni-runner.js';
 import { openGlobalDb } from './v5/global-db.js';
 import { enqueueApproval, getApproval, listInbox } from './v5/omni-queue.js';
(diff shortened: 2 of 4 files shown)


namastex888:
Superseded: reopened against dev as the base (PRs to main must come from dev). The 4 commits are cherry-picked cleanly onto dev in the new PR.

PR #2591, opened 2026-07-14
fix(agent-sync): honor home publish platform seam

namastex888:
## Context

Carries the two accepted CodeRabbit changes from #2590 that were pushed after GitHub had already merged that PR's earlier head. This branch is rebased onto `dev` 5.260714.7 and contains exactly one commit.

## Changes

- Use the shared explicit no-clobber platform selector for missing-`GENIE_HOME` publication as well as regular-file publication.
- Extract the duplicated Darwin/Linux home publisher result/error handling into `makeAgentSyncHomePublisher()`.
- Add a deterministic forced-Linux test proving unavailable injected native setup selects the dedicated portable `mkdir` commit without process-global platform mutation or an explicit probe object.
- Update the seam documentation to cover both native no-clobber callers.

## Validation

- Affected agent-sync fan-out/lock suite — 65 pass
- Uninstall suite — 74 pass
- Rebased focused regression — pass
- `bun run check:fast` — pass
- `git diff --check origin/dev...HEAD` — pass
- Independent incremental review — SHIP, no CRITICAL/HIGH gaps
- CodeRabbit confirmed both findings addressed; its Darwin caching suggestion was withdrawn after the one-shot workload/lifetime contract was explained.

### DIFF
--- a/src/lib/agent-sync.test.ts
+++ b/src/lib/agent-sync.test.ts
@@ -2753,6 +2753,33 @@ describe('cross-process sync lock', () => {
     expect(readdirSync(absentHome)).toEqual([]);
   });
 
+  test('an explicit Linux home platform with unavailable native setup selects the portable mkdir commit', () => {
+    const absentHome = join(fixture.root, 'linux-native-unavailable-home');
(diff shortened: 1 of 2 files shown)


namastex888:
@coderabbitai review

PR #2593, opened 2026-07-17
feat(methodology): make simplicity a lifecycle gate

lirazsiri:
## What changed

- Make KISS/YAGNI a hard gate across brainstorm, wish, review, fix, work, architecture, dream, and PM guidance.
- Add a required Simplicity Case to design and wish templates.
- Route overdesigned plans back to brainstorm/wish instead of spending fix retries defending optional machinery.
- Keep canonical and shipped plugin skill copies in sync.
- Add a release-doc contract test covering the new lifecycle behavior.

## Why

Genie already valued simplicity, but the lifecycle could still approve speculative caches, coordination states, configuration, or abstractions without a present requirement. This change makes the smallest complete design the baseline and requires added machinery to earn its carrying cost with current evidence.

## Impact

Plans should become easier to review, execute, and maintain. Plausible future complexity is deferred behind measurable adoption triggers instead of being shipped dormant.

## Validation

- `bun install --frozen-lockfile`
- `bun test scripts/release-docs.test.ts` — 18 passed
- `bun run check` — typecheck, Biome, knip, skill/wish/complexity/council/hook/plugin gates passed; the full test phase reached 1,683 passing tests and 4 failures
- The same 4 failing `src/lib/agent-sync.test.ts` cases reproduce on a clean detached `origin/main`: one council journal recovery case and three frozen Codex fallback digest cases. They are upstream baseline failures and are unrelated to this documentation/methodology diff.

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -26,6 +26,7 @@ Genie v5 is zero-daemon except for the explicitly launched `genie omni serve` br
 
 ## Engineering rules
 
+- KISS is a release gate, not a preference. Start with the simplest complete design that satisfies current user stories. Caches, deltas, sharding, background coordination, configurable policy, and other stateful machinery require a present contractual need or measured threshold; hypothetical future scale is not evidence. Prefer bounding data and separating history from current state before adding synchronization protocols.
 - Define type and error boundaries before implementation.
 - Preserve user-owned config and unrelated dirty-worktree changes.
 - Config migrations are narrow, backup-first, idempotent, and covered by fixtures.
--- a/plugins/genie/references/review-criteria.md
+++ b/plugins/genie/references/review-criteria.md
@@ -44,6 +44,13 @@ For PASS criteria, evidence includes:
 - No dead code
 - No unresolved TODOs
 
+### Simplicity
+- Simplest complete design is stated
+- Every cache, delta, shard, queue, retry state machine, abstraction, and configuration surface has a current criterion or measurement
+- Current state is bounded and history is separated before synchronization machinery is added
+- Plausible future complexity is deferred behind a concrete adoption trigger
+- Repeated gaps in optional machinery route back to planning instead of consuming fix loops
+
 ### Performance
 - No obvious N+1 queries
 - No unnecessary loops/allocations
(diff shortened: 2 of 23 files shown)


automagik-genie:
Thanks @lirazsiri — the Simplicity Case idea is genuinely good. Right now `main` has exactly one line on this ("YAGNI and simplicity first" in the brainstorm skill), and turning it into a real lifecycle gate with a template section and an `overdesigned-plan` route is the kind of thing that pays for itself. I'd like to take it.

One thing to split out first: the diff also raises the max fix-loop budget from 2 to 3 (in `fix/SKILL.md` frontmatter, the verdict table, the "never exceed N fix loops" rule, the diagnosis format, and `review/SKILL.md`), and the new contract test pins it with `expect(fix).toContain('up to 3 loops')`. That isn't mentioned in the description, and it's a real cost change — each loop is a fixer plus a reviewer dispatch. Could you drop it from this PR and, if you think 3 is right, open it separately with the reasoning? It's worth its own discussion rather than riding along.

Also, since you opened this we moved to a dev-first flow: work lands on `dev` first and `main` gets a twin PR afterward. So this needs a rebase onto `dev` and a retarget. Good news is the base is 383 commits behind but I test-merged it and there are no conflicts — only 3 of your 23 files were touched upstream.

On the Gemini suggestion about asserting plugin-mirror parity in the test: you can skip it, `bun run check` already has a plugin-parity gate covering that.

automagik-genie:
The Simplicity Gate landed on dev as #2698, with your commit preserved exactly as authored plus two follow-up commits per the triage note: the fix-loop budgets are restored to their current values (2 review loops / 1 quality loop — the diff also touched `plugins/genie/references/review-criteria.md` and the quality-review budget in `work/SKILL.md`, broader than first flagged), and the escalation-diagnosis sentence now says "rules out the other four causes" since `overdesigned-plan` makes five. If you believe 3 loops is the right budget, please propose it standalone with the reasoning — it deserves its own review rather than riding along.

Closing in favor of #2698. This was the best of your three PRs — thanks @lirazsiri, the `overdesigned-plan` escalation route is a genuinely good addition to the lifecycle.

PR #2594, opened 2026-07-17
feat(orchestration): isolate concurrent delivery lanes

lirazsiri:
## What changed

- Define `genie launch` worktrees as the isolation boundary for parallel delivery lanes.
- Clarify that task claims coordinate shared-workspace scope and do not imply filesystem isolation.
- Align Claude and Codex role guidance, lifecycle skills, and runtime-neutral dispatch documentation with that contract.
- Extend init, launch, and release-documentation tests to enforce the policy.

## Why

Native subagents can share a workspace, so concurrent writers need an explicit isolation contract. This revision makes the existing worktree mechanism the canonical boundary and documents when agents must sequence work instead.

## Impact

Parallel implementation groups get isolated Git worktrees through `genie launch`; reviewers and scouts remain read-only, and shared-workspace work stays coordinated through visible task claims.

## Stack

Depends on #2593 (`feat(methodology): make simplicity a lifecycle gate`). Until that PR merges, GitHub shows both commits in this PR; afterward this PR reduces to commit `a82507b3`.

## Validation

- `bun run check:fast` — passed in the pre-push hook.
- `bun test src/term-commands/init.test.ts src/term-commands/launch.test.ts scripts/release-docs.test.ts` — 71 passed, 0 failed.
- `bun run check` — reached the full suite; 1,683 passed and 6 failed. The same six failures reproduce on parent commit `28377f9e`, so they are inherited from #2593 and not introduced by this revision.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -72,6 +72,9 @@ yarn-error.log*
 .genie/genie.db-wal
 .genie/genie.db-shm
 
+# Per-worktree kickoff prompts emitted by `genie launch`
+.genie/launch/
+
 # Local genie state (not part of the package)
 .genie/agents.json
 .genie/teams/
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -26,6 +26,7 @@ Genie v5 is zero-daemon except for the explicitly launched `genie omni serve` br
 
 ## Engineering rules
 
+- KISS is a release gate, not a preference. Start with the simplest complete design that satisfies current user stories. Caches, deltas, sharding, background coordination, configurable policy, and other stateful machinery require a present contractual need or measured threshold; hypothetical future scale is not evidence. Prefer bounding data and separating history from current state before adding synchronization protocols.
 - Define type and error boundaries before implementation.
 - Preserve user-owned config and unrelated dirty-worktree changes.
 - Config migrations are narrow, backup-first, idempotent, and covered by fixtures.
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -32,7 +32,16 @@ Genie v5 is zero-daemon except for the explicitly launched `genie omni serve` br
 - Config migrations are narrow, backup-first, idempotent, and covered by fixtures.
 - Every new CLI surface tests success, error exit code, stderr, and idempotency.
 - Shared skills use roles and native delegation language, never a hardcoded client tool name.
-- Subagents share a workspace unless the client explicitly guarantees otherwise. Genie task claims own shared-workspace scope; `genie launch` owns worktree isolation.
+- Treat each concurrently active execution group as an isolated delivery lane with its own branch and worktree; never
+  run parallel writers in one checkout, even when their expected file scopes are disjoint. The PM owns the wish
+  integration worktree, merge order, and conflict resolution. After a reviewed group is merged and validated, the PM
+  removes its clean worktree and merged branch so active worktrees and branches represent active work.
+- Mainline ownership follows repository topology. When `main` has an authoritative GitHub upstream, local `main` is a
+  clean fast-forward mirror of that remote's `main`: wishes reach it through reviewed GitHub PRs, never a local feature
+  merge. With no configured remotes, the PM may integrate a finished wish through a validated temporary candidate,
+  archive that exact integrated closure as `archive/wish/<slug>`, remove its clean active lanes, and then fast-forward
+  unchanged local `main` to the archived commit. A non-GitHub or ambiguous remote requires an explicit user-selected
+  integration policy.
 - Reviewer and engineer are different roles. Never accept self-review as independent evidence.
 - Codex agents inherit the active model; do not hardcode unstable model identifiers.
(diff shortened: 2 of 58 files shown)


automagik-genie:
Thanks @lirazsiri, and sorry for the slow response on this one — it's the most substantial of your three PRs and it deserves a real answer rather than a quick merge.

First, you found a genuine bug. `genie launch` writes `.genie/launch/<group>.prompt` into every worktree it creates (`src/term-commands/launch.ts:410`), and `.genie/launch/` isn't in `.gitignore` — so every launched worktree shows up dirty. Your `.gitignore` and `init.ts` changes fix that, and it also happens to be a precondition for the cleanliness check in your own GC step. We'd like to land that part on its own, as a small standalone PR, so it isn't gated on the larger discussion below. Happy for you to open it, or we can carve it out and credit you.

On the orchestration policy itself, we want to escalate rather than decide it in review, because it reverses a position `main` currently holds deliberately. `AGENTS.md:34` today says subagents share a workspace and that task claims own shared-workspace scope while `genie launch` owns worktree isolation, and `skills/work/SKILL.md:43` allows "disjoint file ownership **or** dedicated worktrees." Your change makes worktrees mandatory even for disjoint scopes. That may well be the right call — shared-workspace races are real — but as the PR notes, when the runtime can't place native subagents in worktrees the rule degrades to sequencing, and that's a throughput tradeoff the maintainer should weigh explicitly. Flagging it for Felipe.

Two more things to fold in when we get there. The mainline-ownership section was written before we moved to a dev-first flow: work now lands on `dev` first and `main` receives a twin PR, so the text describing `main` as a fast-forward mirror fed by reviewed PRs needs reworking against that. And Gemini's note on the GC snippet is correct — `test -z "$(git -C <group-worktree> status --porcelain)"` succeeds when the directory is missing or git errors, so the cleanliness gate fails open right before a worktree removal. `status_output=$(git -C <group-worktree> status --porcelain) && test -z "$status_output"` closes it.

Since this touches `init.ts` and `launch.ts`, it also needs our twin-PR choreography (human merges the `main` PR, then the `dev` twin), which is another reason to keep the `.gitignore` fix separate. Leaving this open pending Felipe's call on the isolation policy.

automagik-genie:
Update: the standalone bug fix from this PR landed on dev as #2697 (reimplemented against current dev, credited via Co-authored-by — thanks for finding it). One extra observation from that work worth knowing: `genie launch` also writes `.mcp.json`, `.warp/` and `.codex/` into every worktree via `registerProjectMcpConfigs`; those are deliberately not gitignored since a repo may want them committed, so ignore-vs-commit for those is a separate product decision.

The orchestration-policy core of this PR remains open for Felipe's decision. For the record, the maintainer-side recommendation under discussion is a middle path: flip the **default** to worktree isolation for parallel writers while keeping an explicit escape hatch when the orchestrator asserts disjoint file scopes — rather than an unconditional mandate that degrades native-subagent dispatch to sequencing. If that lands, it would be re-scoped on top of current dev (the mainline-ownership section predates the dev-first flow) and would fold in the GC cleanliness fix from the review thread.

automagik-genie:
Closing in favor of the landed adaptation — and the full story is worth telling, because this PR ended up shaping the repo's concurrency policy more than most merged ones do.

**What happened to it:** your isolation thesis triggered a formal 4-lens council deliberation (2026-07-27), which surfaced that the recorded context-mixing incident was a *repo-level git-state mutation* (a concurrent session switching the shared checkout), not a file-scope collision. That reframing produced wish `worktree-isolation-hardening`, plan-reviewed through 2 fix loops (including a live probe experiment on native worktree isolation), executed in 3 groups, each independently reviewed, and merged to dev as #2707. Every commit carries `Co-authored-by: Liraz Siri`.

**What landed from your design:**
- The **git-state freeze**: shared-workspace subagents never `checkout`/`switch`/`reset`/`stash`/`rebase` — only the orchestrator moves HEAD. Your isolation instinct, aimed at the invariant the incident actually violated.
- Your **reviewer-snapshot design**: reviews of committed work now pin to a detached read-only worktree at the exact reviewed commit.
- Your **GC design**, hardened: `genie doctor --fix` now removes merged+clean launch worktrees fail-closed in code (ancestry + cleanliness proofs chained; review even caught and regression-locked a tag-shadowing attack on the ancestry proof before landing). Your `.gitignore` fix had already landed via #2697.

**What was deliberately deferred, with recorded triggers** (AGENTS.md "Flip conditions", linked issues #2705/#2706): the unconditional mandate flips ON automatically-by-policy when the native isolation guard's ergonomics mature (one gap left — a live pilot is filed), or when evidence shows the freeze insufficient; your PM integration-worktree protocol lands on the first orphaned-lane or wrong-order-merge incident. So the parts not adopted today are not rejected — they have named activation conditions in durable policy text.

The mainline-ownership section predates the dev-first flow and stays out. Thanks for a PR that was worth this much machinery — the diagnosis was right even where the prescription was resized.

PR #2595, opened 2026-07-19
docs: teach reader-boundary documentation

lirazsiri:
## What changed

- teach the docs skill to explain decisions, actions, observable outcomes, failure behavior, and next steps
- keep implementation machinery out of user-facing documentation unless it is needed for safe use, troubleshooting, or extension
- add the same mechanism-leakage guidance to the DX documentation audit skill
- keep canonical skills and plugin mirrors synchronized

## Why

User documentation should preserve the reader's attention for what they need to know. When an internal mechanism does not affect a decision or interaction, documentation can promise the observable result instead of requiring a detailed mental model.

## Validation

- push-time static check suite passed, including typecheck, Biome, dead-code analysis, skill and wish lint, complexity budget, workflow checks, and bundled-hook parity
- canonical and plugin skill mirrors are byte-identical
- full test suite currently has six unrelated runtime failures that reproduce on main

### DIFF
--- a/plugins/genie/skills/docs/SKILL.md
+++ b/plugins/genie/skills/docs/SKILL.md
@@ -43,4 +43,5 @@ Example brief: "Audit README.md, CLAUDE.md, and skills/work/SKILL.md after PR #7
 ## Rules
 - Grounded progress: report only what was audited or generated in this session, each claim backed by a check actually run — "3 files verified current, 1 updated, 0 dead references", never just "docs written".
 - No fiction: never document features that don't exist yet; no dead paths or APIs.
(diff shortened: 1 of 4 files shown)


automagik-genie:
Thanks for this, @lirazsiri — the reader-boundary framing is a real gap in our docs guidance, and I checked: nothing equivalent has landed on `main` in the meantime. The two bullets still apply cleanly to the current files.

One process note: since you opened this, we moved to a dev-first flow — changes land on `dev` first, and `main` is the control plane that receives a twin PR afterward. This PR targets `main` directly, so it needs a quick retarget. The base is also about 383 commits behind now, though I test-merged it and there are no conflicts.

Could you rebase onto `dev` and retarget the PR there? Once it's green on `dev` we'll handle the `main` twin on our side. Happy to do the rebase ourselves if you'd rather — just say the word.

automagik-genie:
Landed on dev as #2696 with your commit authorship preserved (cherry-pick applied clean — zero conflicts despite the base drift). Thanks @lirazsiri! Closing this original in favor of the landed version per the dev-first flow; it will reach main with the next rolling promotion.

PR #2716, opened 2026-07-27
fix(release): refuse re-dispatch of published versions + delete dead channel default (#2674, #2675)

### LINKED ISSUE #2674: Release ops rule: never re-dispatch a version that already has a published release (pre-fanout releases are non-replayable)
From the publish-tail review before the first post-fanout release (no blockers found for fresh versions; verdict SHIP).

**Rule: never re-dispatch Release for a version that already has a published GitHub Release. Always allocate a fresh version.** Both latent defects below fire only on that precondition, and neither can fire on a fresh tag (version.yml always allocates max-plus-one and pushes a fresh tag).

**F1 — asset-inventory expansion makes pre-fanout releases un-replayable.** scripts/reconcile-release-assets.sh now expects 20 (dev) / 28 (homolog) / 36 (stable) assets; every published release has exactly 12, draft=false, immutable=true. A re-dispatch of e.g. v5.260724.1 sees an incomplete published immutable release and exits 3 — unrecoverable, because immutability means the missing descriptor/endorsement assets can never be added.

**F2 — released_at derivation changed.** reconcile-channel-manifests.sh now derives a deterministic midnight-UTC released_at from the version date, and equal-version replays require exact byte match. Manifests on main carry wall-clock timestamps (latest/homolog: 2026-07-21T00:13:26Z; dev: 2026-07-23T06:43:34Z), so re-releasing v5.260720.10 or v5.260723.8 fails cmp in the manifests job AFTER the release is published and locked.

Suggested hardening beyond the rule: teach the publish admit path to refuse a version whose tag already carries a published release, so the rule is enforced mechanically instead of remembered.

### LINKED ISSUE #2675: release-publish.yml: delete the dead CHANNEL=stable default in publish; pin multi-attestation and CAS-loop coverage
Two INFO findings + test gaps from the publish-tail review (no blockers).

**Dead stable default (delete it).** release-publish.yml publish job: `CHANNEL="${INPUT_CHANNEL:-}"; [[ -z $CHANNEL ]] && CHANNEL=stable` is unreachable — admit exits 1 on an empty/unknown channel and both callers always pass one — but it is inconsistent: publish uses steps.meta.outputs.channel while finalize/manifests use inputs.channel raw. If it ever fired, publish would upload a 36-asset stable inventory while finalize aborts, leaving a published-but-unfinalized release. Fail hard like admit instead of defaulting.

**Test gaps (no test would catch):**
- The attestations API returns TWO attestations per published tarball (GitHub's in-toto release/v0.2 immutable-release attestation + our sign-attest one). Production code tolerates it (first()/any() selects by predicate type) but the fake gh in reconcile-release-assets.test.ts always emits exactly one — pin the multi-attestation shape.
- The manifests job's inline CAS loop (fetch/detach/reconcile/add/commit/push with failure classification) has no test; only reconcile-channel-manifests.sh itself is tested.
- The publish job's sed -E version parse and the channel default above are untested workflow-inline shell.

automagik-genie:
> **TWIN-PR — this is the `main` half. Human-merge only.**
> This touches `.github/workflows/release-publish.yml`, the live release surface. **Do NOT merge while a stable release run is in flight** (a run that has passed `admit` reads its control code from the admitted `main` commit; merging under it changes the workflow mid-flight). Merge only when no release run is executing, and land the `dev` twin alongside it so the two branches do not diverge on this file.

Closes #2674. Closes #2675.

### 1. `admit` refuses re-dispatch of an already published version (#2674)

New step `Refuse re-dispatch of an already published release`, running **after** the caller-binding step (an unproven caller is still rejected before any API call):

- `gh api repos/<repo>/releases/tags/v<version>` with `GH_TOKEN: ${{ github.token }}`.
- **Published (`draft == false`) → job fails** with `::error ::release-replay.published`, citing the rule ("published releases are non-replayable — allocate a fresh version") and #2674.
- **Draft → admitted**, on purpose: an interrupted run must still be able to finish the release it already opened. `reconcile-release-assets.sh` resumes partial drafts by design.
- **404 → admitted** (fresh version, the normal path).
- **Any other API answer → fails closed** (`::error ::release-replay.unknown`) instead of silently admitting on an ambiguous read.

Why mechanically and not by convention — both defects from the publish-tail review fire only on this precondition and both land *after* the release is published and locked, i.e. unrecoverable:

- **F1** `scripts/reconcile-release-assets.sh` now expects the 20 (dev) / 28 (homolog) / 36 (stable) delivery fanout; every pre-fanout published release carries exactly 12 assets and is `immutable=true`, so the missing descriptors can never be added → `exit 3`.
- **F2** `scripts/reconcile-channel-manifests.sh` now derives a deterministic midnight-UTC `released_at`; the manifests already committed to `main` carry wall-clock timestamps, so an equal-version replay fails `cmp` in the manifests job.

`version.yml` always allocates max-plus-one and pushes a fresh tag, so no legitimate release path is affected.

**Permissions note:** `admit` moves from `permissions: {}` to `permissions:\n  contents: read`. That is the least the query needs, it matches the workflow-level default, and `contents: write` still starts only at the `publish` job — nothing may mutate the repository before the caller is proven. `sign-attest.yml`'s admit keeps `permissions: {}`.

### 2. Dead `CHANNEL=stable` default deleted from `publish` (#2675)

`publish`'s `Resolve version + channel + asset inventory` step carried:

```bash
# Channel: workflow_dispatch picks; workflow_run defaults to stable.
CHANNEL="${INPUT_CHANNEL:-}"
if [[ -z "${CHANNEL}" ]]; then CHANNEL="stable"; fi
```

Unreachable (`admit` exits 1 on an empty/unknown channel and both callers always pass one), the comment was stale (this workflow is `workflow_call`-only), and it was the one place reading `steps.meta.outputs.channel` while `finalize` and `manifests` read `inputs.channel` raw. Had it ever fired, `publish` would have uploaded a 36-asset stable inventory that `finalize` then aborts on — a published-but-unfinalized release.

Now: the step is `Resolve version + asset inventory` (version parse + upstream cross-check only, unchanged), and `publish` reads `CHANNEL: ${{ inputs.channel }}` exactly like `finalize`/`manifests`. Fail-hard on a bad channel stays where it belongs, in `admit`.

## Tests

`scripts/release-docs.test.ts` gains two pinning tests (both branches):

- `admit refuses re-dispatch of a version that already has a published release` — pins the guard, the `.draft` check, the fail-closed branch, the issue citation, and that the guard runs after caller binding.
- `publish reads the channel input directly instead of defaulting it` — pins that the publish job carries no `INPUT_CHANNEL` / `CHANNEL="stable"` / `steps.meta.outputs.channel`, and that publish/finalize/manifests all read `inputs.channel`.

The existing `permissions: {}` assertion was split: `sign-attest.yml` still asserts it; `release-publish.yml`'s admit block now asserts `contents: read` and explicitly *not* `contents: write` / `id-token:`.

**Multi-attestation coverage (#2675) rides the dev twin** — `scripts/reconcile-release-assets.test.ts` is unchanged here so the two branches stay identical on the release-critical surface. See the dev PR.

**CAS-loop coverage: not added, recorded here instead.** The manifests job's fetch/detach/reconcile/add/commit/push loop with 403-vs-non-fast-forward failure classification is **workflow-inline shell with no testable seam** — there is no script under `scripts/` that owns it (`reconcile-channel-manifests.sh`, which *is* tested, is only the reconcile step the loop calls). Adding coverage would mean extracting the classifier into a new script, wiring it through the `RUNNER_TEMP` copy the loop already does for the reconciler, and doing so identically on both branches — a substantive refactor of the most fragile job in the release path, out of scope for a hardening PR that must be merged by hand. Per issue #2675 this is deliberately deferred rather than faked.

## Validation

```
bun test scripts/reconcile-release-assets.test.ts scripts/release-docs.test.ts   # 54 pass, 0 fail
bun run check:fast                                                              # OK: all checks passed
bunx actionlint .github/workflows/release-publish.yml                           # only the pre-existing
                                                                                # `queue: max` syntax note,
                                                                                # identical on origin/main
```

`actionlint` ran with `shellcheck` present, so the new inline shell was shellcheck-clean.

---

**Dev twin:** https://github.com/automagik-dev/genie/pull/2717

## Summary by CodeRabbit

* **Bug Fixes**
  * Strengthened release publishing safeguards to prevent reusing versions that already have a published release.
  * Improved release channel handling so publishing consistently uses the selected workflow channel.
  * Restricted release admission workflow access to read-only repository content.

* **Tests**
  * Expanded release workflow validation to cover permission settings, duplicate-release refusal, failure-safe behavior, and channel selection.

### DIFF
--- a/.github/workflows/release-publish.yml
+++ b/.github/workflows/release-publish.yml
@@ -82,7 +82,12 @@ jobs:
     name: Admit trusted release orchestrator
     runs-on: ubuntu-latest
     timeout-minutes: 2
-    permissions: {}
+    permissions:
+      # Read-only, and only so the replay guard below can ask whether the tag
+      # already carries a published release. contents:write still starts at the
+      # publish job: nothing may mutate the repository before the caller is
+      # proven, and the guard runs after the binding step for the same reason.
+      contents: read
     steps:
       - name: Bind caller workflow and control commit
         shell: bash
@@ -137,6 +142,43 @@ jobs:
               ;;
           esac
           echo "Trusted release caller admitted at ${CALLER_SHA}"
+
+      - name: Refuse re-dispatch of an already published release
+        shell: bash
+        env:
+          GH_TOKEN: ${{ github.token }}
+          RELEASE_REPOSITORY: ${{ github.repository }}
+          INPUT_VERSION: ${{ inputs.version }}
+        run: |
+          set -euo pipefail
+          # A published (non-draft) release is immutable, and this pipeline is
+          # no longer able to replay one: reconcile-release-assets.sh expects
+          # the current 20/28/36-asset delivery fanout while every pre-fanout
+          # release carries 12 that can never be extended, and
+          # reconcile-channel-manifests.sh now derives a deterministic
+          # midnight-UTC released_at that no longer cmp-matches the wall-clock
+          # timestamps already committed to main. Both failures land AFTER the
+          # release is published and locked, so refuse the run here instead.
+          # Always allocate a fresh version (version.yml allocates max-plus-one
+          # and pushes a fresh tag). See automagik-dev/genie#2674.
+          # Drafts stay admissible on purpose: an interrupted run must still be
+          # able to finish the release it already opened.
+          RELEASE_JSON="$(mktemp)"
+          RELEASE_ERROR="$(mktemp)"
+          if gh api "repos/${RELEASE_REPOSITORY}/releases/tags/v${INPUT_VERSION}" \
+            >"$RELEASE_JSON" 2>"$RELEASE_ERROR"; then
+            if [[ "$(jq -r '.draft' "$RELEASE_JSON")" != true ]]; then
+              echo "::error ::release-replay.published v${INPUT_VERSION} already has a published GitHub Release — published releases are non-replayable (immutable asset inventory + deterministic released_at); allocate a fresh version instead. See automagik-dev/genie#2674"
+              exit 1
+            fi
+            echo "v${INPUT_VERSION} exists only as a draft release; admitting so the interrupted publish can complete"
+          elif grep -q 'HTTP 404' "$RELEASE_ERROR"; then
+            echo "v${INPUT_VERSION} has no GitHub Release yet"
+          else
+            cat "$RELEASE_ERROR" >&2
+            echo "::error ::release-replay.unknown could not determine whether v${INPUT_VERSION} already has a published release; failing closed. See automagik-dev/genie#2674"
+            exit 1
+          fi
 
   prepare-delivery-evidence:
     name: Build canonical delivery descriptors
@@ -1012,12 +1054,15 @@ jobs:
           name: delivery-candidate-manifests
           path: ${{ runner.temp }}/candidate-manifests
 
-      - name: Resolve version + channel + asset inventory
+      # Channel is NOT resolved here: admit already rejects an empty/unknown
+      # channel, and finalize/manifests read inputs.channel raw. A local
+      # default would only let publish upload a stable-sized inventory that
+      # finalize then aborts on, leaving a published-but-unfinalized release.
+      - name: Resolve version + asset inventory
         id: meta
         shell: bash
         env:
           INPUT_VERSION: ${{ inputs.version }}
-          INPUT_CHANNEL: ${{ inputs.channel }}
         run: |
           set -euo pipefail
           cd dist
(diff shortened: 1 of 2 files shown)


automagik-genie:
Superseded: identical content reached dev via twin #2717 (merged) and rides the consolidated dev→main promotion; a separate main-branch PR would double-apply the diff. #2674/#2675 close when the promotion lands.

PR #2723, opened 2026-07-28
fix(router): bypass Genie when lifecycle adds no value

lirazsiri:
## What changed

- add a lightweight value gate before Genie's lifecycle intent router
- preserve Genie routing for explicit requests and work related to an existing wish or brainstorm
- default unrelated ordinary work to the host agent workflow
- use cheap read-only inspection when lifecycle value is unclear
- keep canonical skills and the shipped plugin mirror byte-identical

## Why

The router currently sends every clear feature, fuzzy request, or bug report into a durable Genie workflow. That imposes substantial token, latency, artifact, and orchestration costs even when ordinary agent execution is sufficient.

The new policy activates Genie only when the user asks for it, related lifecycle state exists, or the work demonstrably needs durable planning, unresolved product or architecture decisions, coordinated workstreams, or tracking across sessions or handoffs. Being a feature, bug, security-sensitive change, or multi-file edit is not enough by itself.

## Impact

Simple and otherwise ordinary requests avoid Genie-specific artifacts, roles, announcements, and review gates. Existing wishes still resume through persisted state, and repository safety, worktree, validation, and review requirements continue to apply outside Genie.

## Validation

- `bun scripts/sync-plugin-skills.ts --check`
- `bun test scripts/release-docs.test.ts scripts/sync-plugin-skills.test.ts` (26 passed)
- `git diff --check`

### DIFF
--- a/plugins/genie/skills/genie/SKILL.md
+++ b/plugins/genie/skills/genie/SKILL.md
@@ -1,13 +1,13 @@
 ---
 name: genie
-description: "Entry point for all genie operations — auto-routes natural language to the right skill, detects lifecycle state, and handles operational commands. Use when planning features, reporting bugs, orchestrating execution, or asking about genie."
+description: "Entry point for Genie operations — resumes existing lifecycle state and orchestrates work that needs durable planning or coordination. Ordinary requests bypass Genie unless the user asks for it."
 ---
 
 # genie — Auto-Router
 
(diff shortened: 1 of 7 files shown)


namastex888:
can you rebase your PR`s to dev pls? :) @lirazsiri 
ps: thanks for the contributions

PR #2768, opened 2026-08-12
fix(task-state): serialize setTaskWish read and no-op check inside the immediate transaction

namastex888:
## Summary
- `setTaskWish` (assign / clear / reassign of a card's wish identity) read the current row, ran the no-op comparison, and derived the timeline note **before** its `apply.immediate()` transaction. With worktrees sharing one `genie.db` across processes, a racing writer could commit in that window, producing a stale `from` in the wish-event note, a stale no-op decision, or a raw FK-constraint error (instead of `UnknownTaskError`) on a concurrent delete.
- The row read, existence check, no-op comparison, and event append now all run inside one immediate transaction, so every writer derives its event from the row state its own transaction serialized against — same pattern as `checkoutTask` / `deleteTask`. `linkTaskToWish` inherits the fix.

## Regression coverage
- New multi-process test (modeled on the roster-race test): 9 concurrent bun processes assign/clear/reassign one card via `Promise.allSettled`, asserting every writer commits cleanly and the wish-event timeline forms an unbroken `from→to` chain ending at the final row identity — exactly the invariant a stale pre-transaction read breaks.

## Validation
- `bun test src/lib/v5/task-state.test.ts`: 99 pass / 0 fail; new race test passed 3/3 repeat runs
- `tsc --noEmit` clean, biome clean on both changed files


namastex888:
Superseded by a dev-based branch: this PR was cut from main, so the merge into dev dragged in main's two MCP write-path commits, and that merge combination (not the task-state fix) is what broke the mcp write-protected test. Replacement PR carries the same fix rebuilt on origin/dev, plus the fix for the actual finding target (assignTask/clearTaskAssignment, which only exist on dev).

PR #2799, opened 2026-08-15
chore(release): promote dev to main — GENIE_HOME umask install hotfix

automagik-genie:
## Summary

Promotes `dev` to `main` to release the fix for the fresh-install failure affecting all Linux machines with umask `002` (Ubuntu/Debian default):

```
InstallPromotionError: GENIE_HOME has unsafe permissions
```

## Contents

- **#2798 — fix(install): create GENIE_HOME with mode 0o700 everywhere it is first created.** Every code path that can first-create `~/.genie` now declares `mode: 0o700` (which no umask can widen), matching the install promoter's `0o022` safety predicate. Includes a two-layer regression guard: a negative source scan over `src/` plus a pinned-site test for parameter-carried paths the scan cannot see, with behavioral tests driving the real creator functions under umask `002`.
- **Review hardening** from CodeRabbit findings: nested umask restores now unwind in reverse order, and `SAFE_MODE` rejects explicit modes carrying group/other write bits (`0o775`/`0o777`).
- **Review follow-ups (710beab62)** — test-only hardening from the PR review: an explicit 30s timeout on the cross-process MCP claim-contention test, `fileURLToPath` for the test's `REPO_ROOT` (percent-encoded checkout paths), and the negative `mkdirSync` scan now matches multi-line/Biome-wrapped calls.
- **e3c14d92d — fix(v5): serialize cross-process WAL-index recovery behind a stale-safe lock.** Root cause of the intermittent claim-contention CI failure: two MCP servers starting concurrently on the same `genie.db` both entered the single-owner WAL-recovery path and raced on handles/WAL-index mapping, intermittently killing one child natively on Linux. Recovery now takes a PID-aware, self-reaping lock for the whole probe/rebuild/reopen sequence; lock-timeout maps to the existing busy semantics.
- **df1e3d34c — test(v5): surface spawned MCP server stderr and exit status** so any future harness failure names its cause.
- `[auto-version]` bumps riding the fixes; currently `5.260816.2`.

## Remediation for already-affected machines

This fix stops new installs from creating a group-writable `~/.genie`; it does not (by design) change the mode of an existing directory. A machine where an earlier failed install already left `~/.genie` at `0775` will still be rejected by the promoter's safety check. One-time remedy:

```bash
chmod 700 ~/.genie
```

then re-run the installer.

## Validation

All 17 CI checks passed on #2798 against `dev` (Linux glibc/musl/arm64 builds, E2E v5 lifecycle, unit gate, plugin smoke, security scans). The intermittent cross-process claim-contention failure that blocked earlier heads (pre-existing on dev since Aug 13–14, unrelated to the umask diff) was traced to the WAL-recovery race fixed by `e3c14d92d`; on `df1e3d34c` both the push and pull_request CI runs are fully green (3347 pass / 0 fail; the contention test passes in ~93ms). Merging this PR lets the release automation advance the stable manifest and unbreak `install.sh` for new users.

## Summary by CodeRabbit

- **Security**
  - Genie home, workspace, backup, log, cache, trust, lease, and database directories now use owner-only permissions (`0700`).
  - Directory creation remains secure even when permissive system umasks are configured.

- **Bug Fixes**
  - Improved protection for newly created application state and diagnostic files.

- **Tests**
  - Added regression coverage for directory permissions across supported creation paths and environments.

- **Chores**
  - Updated Genie package and plugin versions to `5.260815.2`.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -10,7 +10,7 @@
   "plugins": [
     {
       "name": "genie",
-      "version": "5.260816.1",
+      "version": "5.260816.2",
       "source": "./plugins/genie",
       "description": "Human-AI partnership for Claude Code. Share a terminal, orchestrate workers, evolve together. Brainstorm ideas, wish them into plans, make with parallel agents, ship as one team. A coding genie that grows with your project."
     }
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -10,7 +10,7 @@
   "plugins": [
     {
       "name": "genie",
-      "version": "5.260815.1",
+      "version": "5.260815.2",
       "source": "./plugins/genie",
       "description": "Human-AI partnership for Claude Code. Share a terminal, orchestrate workers, evolve together. Brainstorm ideas, wish them into plans, make with parallel agents, ship as one team. A coding genie that grows with your project."
     }
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -10,7 +10,7 @@
   "plugins": [
     {
       "name": "genie",
-      "version": "5.260814.4",
+      "version": "5.260815.1",
       "source": "./plugins/genie",
       "description": "Human-AI partnership for Claude Code. Share a terminal, orchestrate workers, evolve together. Brainstorm ideas, wish them into plans, make with parallel agents, ship as one team. A coding genie that grows with your project."
     }
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -10,7 +10,7 @@
   "plugins": [
     {
       "name": "genie",
-      "version": "5.260815.2",
+      "version": "5.260816.1",
       "source": "./plugins/genie",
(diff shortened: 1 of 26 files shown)


automagik-genie:
## Release review — dev → main

**Verdict: SHIP.**

- **Scope verified:** the diff is exactly the GENIE_HOME umask hotfix (#2798, 15 source/test files) plus the `[auto-version]` bump to **5.260815.1** across the eight version manifests (`package.json`, plugin manifests, `marketplace.json`) — all pure version-string changes, nothing else rides along.
- **Fix verified:** every first-creator of `GENIE_HOME` declares `mode: 0o700`; the negative source scan plus pinned-site tests guard the class; behavioral tests assert the resulting mode against the promoter's `0o022` predicate under umask `002` — the exact configuration that reproduced the field failure.
- **Review findings addressed:** nested umask restores unwind in reverse order; `SAFE_MODE` now rejects explicit modes with group/other write bits. The remaining CodeRabbit suggestion (moving the workspace fixture out of `homedir()`) was deliberately declined: `saveWorkspaceRoot` refuses tmpdir roots by design, the fixture is `mkdtemp`-scoped and cleaned up, and restructuring production code to accommodate the test isn't warranted.
- **CI:** all 17 checks passed on #2798 against `dev` at the same tree.

Merging this releases the fix that unbreaks `install.sh` for every fresh Linux install with umask `002`.

namastex888 (approved):

PR #2961, opened 2026-09-18
fix(skills): resolve real paths in design-review-evidence entry-point guard

namastex888:
## Contract

**Core:** The design-review-evidence gate refuses an evidence-less DESIGN.md when the script is invoked through a symlinked path, exactly as it already does when invoked by its real path — i.e. the entry-point guard compares real paths on both sides, in both byte-identical copies of the script.

**Oracle:** `bun test scripts/design-review-evidence.test.ts` — specifically the new regression test that spawns `node <symlinked references dir>/design-review-evidence.mjs verify <DESIGN.md with no evidence block>` and expects exitCode 1 with stderr containing `DESIGN.md must contain exactly one bounded design-review evidence block`; that same test must fail on the pre-fix script (exit 0, empty stderr), and the existing byte-identity test in the same file must still pass.

**Declared files:**
- `skills/wish/references/design-review-evidence.mjs`
- `skills/brainstorm/references/design-review-evidence.mjs`
- `scripts/design-review-evidence.test.ts`

## Acceptance criteria

- In `skills/wish/references/design-review-evidence.mjs`, the entry-point guard no longer compares a bare `resolve(process.argv[1])` against a bare `fileURLToPath(import.meta.url)`: the diff shows BOTH sides passed through realpath resolution before comparison. Falsified if either side of the equality is still the unresolved value, or if only argv[1] is realpath'd while the module path is not.
- The argv[1] side tolerates a path that does not exist on disk — the realpath call is wrapped so that a throw (e.g. ENOENT) falls back to the plain resolved path rather than crashing or silently skipping the CLI body. Falsified if `realpathSync` is called on argv[1] with no try/catch or equivalent guard, or if the fallback returns a value that can never match the module path.
- `realpathSync` is added to the existing `node:fs` import statement in `skills/wish/references/design-review-evidence.mjs` (previously `import { readFileSync, writeFileSync } from 'node:fs';`). Falsified if `realpathSync` is used without being imported, or imported via a second redundant `node:fs` import statement.
- `skills/brainstorm/references/design-review-evidence.mjs` receives the byte-identical same change: the diff hunks for the two .mjs files are identical in content (same added/removed lines, same context), so the two files remain byte-for-byte equal after the commit. Falsified if the brainstorm copy is missing from the diff, or if its hunks differ in any character (including whitespace or comment text).
- `scripts/design-review-evidence.test.ts` gains exactly one new test (no existing test deleted or weakened) that (a) creates a DESIGN.md containing no design-review evidence block, (b) creates a symlink to the directory holding the script under test, and (c) spawns the script through that symlinked directory path with the `verify` subcommand. Falsified if the new test invokes the script by its real path, if it copies the file into place instead of linking, or if it reaches the CLI by a route that does not exercise the symlinked-path entry point.
- That new test asserts `exitCode === 1` AND asserts the refusal text `DESIGN.md must contain exactly one bounded design-review evidence block` appears in the captured stderr. Falsified if it asserts only a non-zero exit, only the message, or a looser match that the pre-fix behavior (exit 0, empty stderr) could also satisfy.
- The new test uses the symlink primitive from `node:fs` (added to the test file's existing `node:fs` import, which currently lacks it) and cleans up its tmpdir the way the neighbouring tests do. Falsified if it writes outside a tmpdir, leaves its symlink or tmpdir behind with no `rmSync`/`afterEach` cleanup, or shells out to `ln` instead of the fs API.
- No file outside the declared set of three paths is modified, added or deleted in the commit. Falsified by any other path in the diff — in particular any `.github/`, `.husky/`, `.claude/hooks/`, `package.json`, `biome.json`, `scripts/release-*` or `release-guard.sh` change.
- The behavioral change is confined to entry-point detection: no change to `reviewableDesign`, to the verify/digest/stamp command implementations, or to any thrown refusal string. Falsified if the diff alters parsing, digesting or stamping logic, or edits the text of the evidence-block error message.

## Gate

```
Ran 2687 tests across 134 files. [389.16s]
```

## Review verdict

SHIP


namastex888:
Closing as the CONTROL arm of the R2 before/after measurement (workflows README, Measured runs): same objective as #2959, run with model opus on every stage and the mikro offload made unavailable. #2959 is the arm that ships. Numbers (phoenix-list-2026-09 price basis): this run $4.77, 42 API calls, 395k subagent tokens, 17.1 min, merge-ready; #2959 $4.33, 52 calls, 412k tokens, 21.6 min, merge-ready. Branch left in place for the record.

PR #2972, opened 2026-09-18
fix(mikro): compute --facts inside the boundary, never on the host

namastex888:
## The defect

`scripts/mikro/call.ts` `runAgent` precomputed `--facts auto` **before** `openBoundary`.

1. **A credentialed call outside the boundary.** `facts.ts` spawned `git` (`:113`) and `gh` (`:122`) with `Bun.spawnSync` and the host environment. Under `--boundary bwrap` that is a host `git` over the tree under `--dir` — which may be the tree under review — and a host `gh` **with the host's GitHub credential**, outside the sandbox and outside the egress ledger. That is exactly the traffic the boundary exists to contain (`EVIDENCE-boundary.md`, `secrets-not-visible`).
2. **A context path the agent could not read.** Facts land in `trustedRoot/.mikro/runs/facts-<runId>.{json,md}`, but `openBoundary` binds only `<dir>/.mikro/runs` (`boundary.ts:502,518`). With `--dir` ≠ the invoking checkout, the agent was handed `context: <path>` (`call.ts:791`) that does not exist inside. Latent, because same-dir runs hide it: `call.test.ts:223` only covered the same-dir case and `facts.test.ts:111` always passed `gh:false`.

## The invariants, and what proves each

| | invariant | proof |
|---|---|---|
| I1 | Under `bwrap` the boundary opens BEFORE the facts, every `git` goes through the session, no `gh` is spawned, and the row says `skipped-boundary` | `call.test.ts` "the boundary opens BEFORE the facts, and every facts command runs inside it" (the fake session's first recorded command is `git`, no `gh` ever, all git before the first `mikro`); `facts.test.ts` "a runner that cannot reach GitHub never issues a gh argv"; live `contained-git-worktree` |
| I2 | `--boundary none` is unchanged apart from the additive `gh` field | `facts.test.ts` "the contained record differs from the `--no-gh` host record in nothing but basis.gh" + "the default runner is the host spawn"; the whole pre-existing suite unchanged |
| I3 | The context file is readable inside when `dir != trustedRoot`, stays under the trusted root, and is NOT writable from inside | live `boundary.test.ts` "a facts file under the trusted root is readable inside the sandbox and cannot be written" (real bwrap, `--dir` ≠ trusted root); pure "a readable path is bound read-only AFTER the writable set, and never joins it" |
| I4 | Boundary-open failure stays fatal; a facts failure stays non-fatal; the boundary closes on every path | unchanged `BoundaryError` propagation + `facts: skipped (…)`; facts now sit INSIDE the existing `try/finally`, and the ordering test asserts `session.closed` after a run that failed |
| I5 | A retry reuses the same facts | the ordering test runs `retries: 1` and asserts 2 `mikro` commands with every `git` before the first of them |

Writable-set assertion in `boundary.test.ts` stays green — the writable set did not grow.

## Decisions taken (no one to ask)

- **Injected runner, not an import.** `buildFacts` takes `runner?: FactsRunner` (default `hostFactsRunner`, the existing spawn). `call.ts` owns `boundaryFactsRunner(session)`. `facts.ts` imports nothing from `boundary.ts`.
- **`gh` under a contained runner is refused in `facts.ts`, not merely unset.** `gh()` short-circuits on `canRunGh: false`, so no code path can reach a host `gh` behind the boundary's back.
- **`basis.gh` lives in the artifact as well as the ledger row** (`host` | `off` | `skipped-boundary`). A reader of `.mikro/runs/<agent>.jsonl` and a reader of the facts file itself both need to tell a *smaller* facts file from a *wrong* one.
- **A contained runner outranks `--no-gh`** when both apply: the boundary is the structural reason, and that is what a reader has to know.
- **No `gh` claim for a caller-supplied `--facts <path>`.** This run did not build that file; the row carries `path` and no `gh` rather than a guess.
- **`--ro-bind-try`, one file, applied after the writable binds.** The facts are written after the preflight, and `argv()` builds a fresh bwrap invocation per command, so the bind must tolerate an absent source at open time and picks up the real file when the runtime is spawned. Read-only on purpose: a facts file is the run's audit record, and an agent that could rewrite it could rewrite its own evidence. Same treatment for a caller-supplied `--facts <path>` (`factsContextPath` is the one place the path is decided, so the bind and the write cannot drift).
- **File READS stay host-side** (`CLAUDE.md`, `AGENTS.md`, `.genie/INDEX.md`, wish/brainstorm titles). The boundary contains execution and credentialed network; `call.ts` already reads the same tree host-side to verify every citation. Named in the README.
- **One test seam:** `RunOptions.openBoundary`. The order is otherwise unprovable without a real bwrap host and a real provider key.

## Evidence

- Focused: `umask 022; bun test scripts/mikro/` → **171 pass, 0 fail, 598 expect()** across 8 files (was 159/554). The two live suites really ran on this host (bwrap 0.11.1, socat, mikro v1.260909.1), verified with `-t "facts context file"` → 2 pass, 35 filtered.
- Full gate: `umask 022; bun run check` → **2837 pass, 3 skip, 0 fail**, 15993 expect(), 140 files. typecheck/lint/skills/wishes/complexity-budget/orca-bundle all clean; the two `noExcessiveCognitiveComplexity` warnings (`doctor.ts:1178`, `orca-orchestration-adapter.ts:789`) and the knip `react`/`dsh-genie-board` notes are pre-existing and in files this PR does not touch.
- Mutation checks (watched each fail for the right reason before trusting it): dropping the `canRunGh` guard fails the no-`gh` test; passing the host runner instead of the contained one fails the ordering test.
- `bun scripts/mikro/boundary.ts --probe` → **12/12**, appended to `EVIDENCE-boundary.md`: the policy control showing the writable set and `secrets-not-visible` are unchanged.
- **One real contained call** (`issue-triage --facts auto --boundary bwrap`, $0.01): `ok: true`, 1 attempt, 53.9 s, ledger `facts: {candidates: 0, ms: 782, gh: "skipped-boundary"}`, egress **4 allowed / 0 denied** — 1 × `api.deepseek.com:443` and **3 × `api.github.com:443`**. Those three are the agent's own `gh`, from inside the sandbox, through the allowlisted proxy: the issue body still reaches the agent, and what used to be one unlogged host call with the host's credential is now three ledger rows. Full round in `EVIDENCE-boundary.md`.

## What I did NOT do

- No `coach`, `triage`, `bench` flags, `wish.js`, or `tags.boundary` on bench runs.
- No refactor of `call.ts` beyond the reordering (facts moved inside the existing `try/finally`).
- Did not touch the usage line at the top of `call.ts` (lines 5–6) — only the `--facts auto` paragraph below it.
- The named residuals stand: credentials are still inside the REPL, and the allowlist is still a host list, not a request policy.
- `--boundary bwrap` + issue mode now yields a much smaller facts file (no issue body to extract keywords from). That is the designed degrade and it is measured above, not guessed; if a round ever shows it costs recall, terminating TLS at the proxy (already the named next slice) is where the issue body comes back.


namastex888:
Stacked on #2971 by the orchestrator: both PRs edit the `call.test.ts` import line and the README ledger-row paragraph, so `wish/mikro-raw-retention` is merged in here (both sides kept; focused `bun test scripts/mikro/` 175 pass / 0 fail, biome + tsc clean) and the base is now that branch. Land #2971 first; GitHub retargets this one to `integration/overnight-0918` automatically. mikro `review-prep` at `c466e4b2c`: ok, 7 claims, 4 flags — the one to keep in mind is that a caller-supplied `--facts <path>` becomes a read-only bind applied last (operator argv, not agent input — accepted).

Issue #2919, opened 2026-09-16
skill-audit-sweep: all-invalid skills list becomes a full sweep; parity signal misreads inside Claude Code; --write not forbidden


namastex888:
## Problems in `.claude/workflows/skill-audit-sweep.js` (dev `1b6ae7e6d`)
1. **An all-invalid `skills` list silently becomes a full sweep** (`:451-453, :468`). `collectNames` logs the dropped entries, but when a supplied list yields no valid name `rosterSource` flips to `inventory` and the whole catalogue is audited. `{skills: ['reveiw']}` launches the expensive full run for an unintended scope. Distinguish "no list" from "a list with no valid entry" and fail the latter.
2. **The parity signal misreads on every in-workflow run.** Inside Claude Code the skills CLI prints an "Agent detected — installing non-interactively" banner and `scripts/skills-inventory-parity.ts` then reports `--list named []`, exit 1, "parser drift". Observed on the 2026-09-16 live run `wf_f4c80be9-388`, which correctly diagnosed it as a false alarm — but every run pays the cost. Reproduce: `bun scripts/skills-inventory-parity.ts` from a Claude Code session; passes with `env -u CLAUDECODE -u CLAUDE_CODE_ENTRYPOINT npx -y skills@1.5.23 add "$PWD" --list > list.txt; bun scripts/skills-inventory-parity.ts --repo . --list-file list.txt`. Either make the parser skip the banner or have the signals prompt run the CI form.
3. **The signals prompt does not forbid `--write`** (`:64-65, :222`) on the inventory-parity script, which rewrites `skills/README.md` under that flag, and names no literal `genie doctor` invocation.

## Evidence
Codex P2 comment on `:453`; independent script audit; live run transcript.

namastex888:
Re-verified 2026-09-19 against dev 22ea62468: items 1 and 3 landed in #2974. Item 2's diagnosis was incomplete — `PARITY_CHECK` at .claude/workflows/skill-audit-sweep.js:63 is the bare `bun scripts/skills-inventory-parity.ts`, which reads stdin (readListInput, :518-523) and can never pass without input; the piped form `npx skills@1.5.23 add "$PWD" --list 2>&1 | bun scripts/skills-inventory-parity.ts --repo .` answers OK for 21 skills. Every sweep still pays the false alarm. Kept open for that one line.

namastex888:
[...]

PR #2975, opened 2026-09-18
fix(skill-audit-sweep): refuse all-invalid skills list and tolerate banner in parity parser

namastex888:
**Contract:** core "The skill-audit sweep refuses a caller `skills` list whose entries are all invalid — with its own named error, raised before any agent is convened, instead of silently becoming an inventory-driven full sweep — while `parseSkillsListOutput` in scripts/skills-inventory-parity.ts parses a `--list` transcript that carries the Claude Code 'Agent detected — installing non-interactively' banner, and the sweep's signals prompt explicitly forbids `--write` on the parity command; each of the three is pinned by an assertion in the two existing parity tests.", oracle "bun test scripts/skill-audit-workflow-parity.test.ts scripts/skills-inventory-parity.test.ts — the two extended parity suites, with the new banner case in describe('parseSkillsListOutput') ('reads the name rows when the skills CLI printed the agent-detection banner') as the focused test for sub-fix 2.", declared files .claude/workflows/skill-audit-sweep.js, scripts/skills-inventory-parity.ts, scripts/skill-audit-workflow-parity.test.ts, scripts/skills-inventory-parity.test.ts

**Gate:** Ran 2832 tests across 140 files. [364.60s]

**Review verdict:** SHIP

**Issue:** 2919


namastex888:
Closed as the CONTROL arm of the R2 Task-lane measurement (same objective, same stage models, mikro offload made unavailable). #2974 is the arm that ships for #2919. Branch kept for the record.

PR #2976, opened 2026-09-18
fix(install): resolve manifest from CDN and GitHub API, pick newer

namastex888:
Contract: core "install.sh's fetch_latest resolves the channel manifest from BOTH the raw CDN and the credential-free GitHub contents API (pinned owner/repo, ?ref=main) and returns the NEWER of the two answers, degrading silently to the CDN answer whenever the API is unreachable, rate-limited or malformed, and saying on stderr when the two sources disagree — pinned by four new cases in the existing install.sh test harness." oracle "bun test scripts/install-swap.test.ts — specifically the new case where a stub PATH curl answers 5.260918.9 for the api.github.com contents URL and 5.260918.1 for the raw.githubusercontent.com URL, and `fetch_latest stable` (sourced under GENIE_INSTALL_SOURCE_ONLY=1) prints the 5.260918.9 payload on stdout and notes the disagreement on stderr. Today this fails because fetch_latest never requests the API URL and prints 5.260918.1." declared files install.sh, scripts/install-swap.test.ts

Acceptance criteria: install.sh gains a second manifest URL for the same channel, pointing at the GitHub contents API for `.well-known/<channel file>` with `?ref=main`, and the channel→file mapping (stable→latest.json, else <channel>.json) is the same rule resolve_manifest_url already uses. FALSIFIED IF: no API URL appears in the diff, or the channel mapping is duplicated with different semantics. | The API host, owner and repo come from pinned literals or the existing REPO constant only — never from manifest payload content (no reading of tarball_base or any manifest field to build a URL) and never from a new environment override. FALSIFIED IF: any part of the API URL is interpolated from a fetched payload or from a newly introduced env var. | fetch_latest issues a GET against BOTH URLs and selects between the two payloads; the selection uses a new numeric, field-wise version comparison (each dotted component compared as a number), not string/lexicographic comparison, and a tie or an unparsable API version keeps the CDN payload. FALSIFIED IF: the diff still shows a single curl GET in fetch_latest, or the comparison is `[[ "$a" > "$b" ]]`/`sort`-free string ordering under which 5.260918.10 would lose to 5.260918.9, or a tie can pick the API copy. | The API read is strictly optional and fails closed to the CDN: a non-zero curl exit (unreachable, timeout, 403/429 under `curl -f`), an empty body, or a body that fails manifest_channel_matches all result in the CDN payload being returned, with no `die`, no `exit`, and no non-zero status propagated out of fetch_latest. FALSIFIED IF: any API-path branch calls die/exit, or a failed API curl is left unguarded (no `|| true` / `|| var=''` / `if` guard) so it can abort the function or the script. | No new hard dependency: the `need curl; need tar; need uname; need ln` list in main() is unchanged, and any tool used only by the API path (base64, jq, python, openssl, …) is guarded by `command -v` with an explicit degrade to the CDN answer when absent. FALSIFIED IF: `need <newtool>` appears, or a binary outside curl/tar/uname/ln and the sed/printf/POSIX builtins install.sh already uses is invoked on the API path without a command -v guard and a CDN fallback. | The API GET carries an explicit bounded `--max-time`, and the existing CDN GET keeps `curl -fsSL` with no newly added time cap. FALSIFIED IF: the API request has no time cap, or the CDN request gains a --max-time (or loses -f/-s/-S/-L) that changes its behaviour on a slow link. | When both sources answer with different versions, the disagreement is reported on stderr through the existing log/warn helpers and names BOTH versions; fetch_latest's stdout still carries nothing but the chosen manifest payload. FALSIFIED IF: the note is printed with a bare `printf`/`echo` to stdout (which would corrupt `payload="$(fetch_latest …)"` at the single call site in main), or the two sources can differ with nothing written to stderr. | Existing CDN semantics are preserved: a CDN payload whose channel does not match still dies with the channel-mismatch message at exit 1, and a CDN read failure with no usable API answer still dies with the `could not fetch` message at exit 5. FALSIFIED IF: either die is removed or its exit code changes. | scripts/install-swap.test.ts gains four cases driven by a stub PATH curl that answers DIFFERENTLY per URL, covering: (a) API newer wins, (b) CDN newer wins, (c) API curl failure degrades to the CDN payload, (d) a malformed/base64-envelope API body degrades to the CDN payload. Each case asserts WHICH payload fetch_latest printed, not merely that it exited 0. FALSIFIED IF: fewer than four such cases exist, the stub curl returns the same body for both URLs, or a case asserts only exit status. | At least one of the new tests asserts the disagreement note is present on stderr with both version strings in it. FALSIFIED IF: no stderr assertion appears in the diff. | The commit changes exactly install.sh and scripts/install-swap.test.ts and no other path; src/genie-commands/install.test.ts in particular is untouched (it overrides fetch_latest wholesale and is insensitive to its internal helpers). FALSIFIED IF: any third file appears in the diff. | The new tests drive install.sh through the existing `GENIE_INSTALL_SOURCE_ONLY=1` source-only seam rather than executing main() or hitting the network. FALSIFIED IF: a new case runs install.sh without that seam, or issues a real request to raw.githubusercontent.com or api.github.com.

Gate: 2831 pass
1 skip
0 fail

Review verdict: SHIP

Issue: 2950

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>


namastex888:
Orchestrator note (mikro `review-prep` at `81aab53e2`, ok, 6 flags — two worth an owner's eye before merge): **(1)** `install.sh` is the shipped `curl | bash` installer and the dual-source read sits directly upstream of artifact selection. Both sources are this repository's `main` (raw CDN vs contents API, owner/repo pinned, no env override) and the chosen manifest still goes through the same signature/attestation verification as before, so "newer wins" can only pick between two published manifests — but this path matches no entry of the mikro BOUNDARIES list and deserves a human read. **(2)** `version_is_newer` fails closed on any non-integer component: an API answer carrying a suffixed version is ignored in favour of the CDN copy (safe direction; not pinned by a test). The operator decision recorded in the run's objective: mirror `fetchLatestManifest` from #2948/#2951 (both sources, newer wins, API strictly optional).

Issue #2918, opened 2026-09-16
research-sweep: conflict gate skips topic overlap, planner may narrow the frozen source list, fence test uses containment


namastex888:
## Problems in `.claude/workflows/research-sweep.js` (dev `1b6ae7e6d`)
1. **Conflict gate skips topic overlap** (`:773-783`, comment at `:273-276`). The comment says every synthesized claim is matched on source, locator AND topic; the conflict branch only requires a quoted finding at the same `{source, locator}`, so a synthesizer-composed conflict can borrow two real but unrelated locators and render as attributed evidence. Fix: add `claimsOverlap(finding.claim, claim)` to the `backing` find at `:780`.
2. **The frozen source list can be narrowed** (`:474-483`). The planner may drop sources; the run only fails when `plannerDrops.length * 2 > kept.length`, so dropping up to half the caller's sources still returns `ok: true` (drops are logged and returned in `droppedSources`, so it is visible, but the front door promises the list is never narrowed). Fix: re-append every omission to a shard, or reject any omission.
3. **Parity test uses containment** (`scripts/research-sweep-workflow-parity.test.ts:48`): `toContain` still passes when `INJECTION_FENCE` loses a clause. Compare the extracted fence with normalized equality.
4. **Stale comment** at `:39-41` claims the parity test pins the `notConvened` labels; it pins the fence, the reader schema and the phase roster.
5. Harmless: the `kept.length >= readersExpected * MIN` guard at `:511` is always true given `:429`.

## Evidence
Independent script audit during the PR #2915 review (read at `1b6ae7e6d`); Codex comments on `:781` and `:479`; CodeRabbit comment on the parity test.

namastex888:
Resolved on dev by #2979 and siblings: the research-sweep conflict gate matches on topic as well as source and locator (claimsOverlap), any planner omission refuses outright, and the fence parity test compares by normalized equality. 6 pass.

PR #2979, opened 2026-09-18
fix(research-sweep): refuse on any planner source omission

namastex888:
## Contract
`.claude/workflows/research-sweep.js` refuses the run when the planner omits even one frozen source (D1: the `plannerDrops.length * 2 > kept.length` threshold is gone, replaced by a `plannerDrops.length > 0` refusal that names every omitted ref and returns the existing `{ok:false, error, droppedSources:[...intake, ...plannerDrops], shards: [], planFallback, planUnderPartitioned, notConvened}` shape, with both prose sites — the comment above the gate and the planner-brief sentence — rewritten to say any omission is refused), the conflict `backing` find also requires `claimsOverlap(finding.claim, claim)` (D2), the parity test's INJECTION_FENCE assertion is normalized equality via the file-local `flat()` (D3), and new script-logic tests pin D1, D2 and D3.

## Oracle
`bun test scripts/research-sweep-workflow-logic.test.ts scripts/research-sweep-workflow-parity.test.ts` — the focused pin is "a single planner omission refuses the run": the lifted gate carries no `* 2 >` threshold and refuses on `plannerDrops.length` alone, returning the ok:false / shards:[] shape; it fails against today's :473, which returns ok:true for one drop out of many.

## Declared files
- `.claude/workflows/research-sweep.js`
- `scripts/research-sweep-workflow-parity.test.ts`
- `scripts/research-sweep-workflow-logic.test.ts`

## Gate
2833 pass, 1 skip, 0 fail

## Review verdict
SHIP

## Issue
#2918


namastex888:
Master-session read of this head (`a2147b0ec`), after `review-prep` (mikro, $0.04, 9 claims, 6 flags):

- **The PR body names D1–D3 only; the diff also carries D4 and D5 of the objective** (issue #2918 items 4 and 5): the stale module comment at `:40-43`, and the dropped conjunct at `:510`.
- **D5 is behaviour-preserving — checked by hand, since no test pins it.** The old guard was `shards.length && shards.length < readersExpected && kept.length >= readersExpected * MIN_SOURCES_PER_READER`, with `readersExpected = max(1, min(maxReaders, floor(kept.length / MIN)))` (`:430`). If `readersExpected >= 2` then `readersExpected <= floor(kept.length / MIN)`, so the third term is true. If `readersExpected == 1` (the only case where the third term CAN be false — a single source), `shards.length < 1` is already false behind the truthy `shards.length`. The term never decided the branch.
- **Flag kept for the owner (D3):** the test makes equality hold by trimming trailing non-bullet lines of the skill section before comparing, so a sentence ADDED to the front door after the last fence bullet would not be compared. Loss of a clause or a bullet — the failure the issue named — is now caught.
- **Flag kept:** the D1/D2 pins lift script text with `new Function`; a refactor that reshapes the gate makes the lift throw rather than fail an assertion — brittle, never a false pass (same idiom as `wish-workflow-logic.test.ts`).
- Operator decision recorded in the run: any `plan.droppedSources` entry naming a frozen ref refuses the run; a source merely assigned to no shard is still appended (it is read, so the list is not narrowed).

Issue #2925, opened 2026-09-16
release-guard rejected v5.260916.5: control listing missed an existing successful main CI push run


namastex888:
## Problem
The dev Release run for `v5.260916.5` (Version run 35113072214, 2026-09-16 15:06Z) failed in `release-guard.sh guard-trusted-release` with:

```
release-guard: trusted main control ancestor 57926b0d9ba7e5ac01044fa17685e7084f2e1cff has no successful CI push run
```

But that run exists: CI `push` on `main` at `57926b0d9`, run 35106197959, completed/success at 14:05:45Z, an hour before the guard ran. The guard's control listing (`gh api repos/<repo>/actions/workflows/ci.yml/runs -f branch=main -f event=push -f status=success -f per_page=100`, `scripts/release-guard.sh:322-323`) did not contain it, so the jq `any(...)` at `:290-302` failed closed. Consequence: the merge of PR #2911 never shipped as a dev release; the tree reaches hosts only when PR #2915 promotes it to stable.

## To investigate
Whether the listing response was truncated, paginated past the run, or served stale by the API; whether `status=success` should be `conclusion`-based; and whether the guard should fall back to `gh api …/actions/runs?head_sha=<control_ci_sha>&event=push` (exact-SHA lookup) before failing. Logs: `gh run view 35113072214 --log-failed`.

namastex888:
Resolved on dev by b9c075a5f: release-guard.sh falls back to an exact head_sha lookup with one retry before failing (scripts/release-guard.sh:284-304, enabled at :430); the failure text now says both the listing and the lookup missed. release-guard.test.ts 43 pass.

Issue #2921, opened 2026-09-16
workfly: overwrite guard rests on the reader's existingNames; model default contradicts the authoring rule


namastex888:
## Problems in `.claude/workflows/workfly.js` (dev `1b6ae7e6d`)
1. **The overwrite guard rests on an agent's directory listing** (`:251-256`, prompt at `:190-196`). The "name already taken" refusal checks only the contract reader's `existingNames[]`; the sole failure check is `!taken.length`. A non-empty but incomplete list lets the author agent overwrite an existing `.claude/workflows/<name>.js` or shadow a skill. Scripts have no filesystem API, so the fix is in the author prompt: "if a file already exists at `<target>`, write nothing and return an empty path", and the script treats an empty `draft.path` as an error. Catalog files are git-tracked, so the exposure is recoverable, but the contract says "returns any collision rather than overwriting anything".
2. **`DEFAULT_MODEL = 'opus'` contradicts the authoring rule.** All four new scripts (`workfly.js:21`, `skill-audit-sweep.js:48`, `docs-audit.js:67`, `research-sweep.js:51`) pass `model: 'opus'` to every agent, while workfly tells the authors it drafts to omit `model` so agents inherit the session model (`:181-183`, `:194`). Decide one rule: default `model` to empty and spread it only when the caller pins one, or document why the catalog pins Opus.
3. `slugify` (`:83`) truncates to four segments, so a five-segment requested name logs a false "the designer renamed …" line (`:257`). `maxRepairs` (`:118`) has no upper cap.

## Evidence
Independent script audit during the PR #2915 review.

namastex888:
**Evidence that this is already fixed on `dev` — ready to close (left open for the owner).**

`bebf0991b` (PR #2932, "real overwrite guard, caller-pinned model only, maxRepairs clamp") is an ancestor of `origin/dev` (`ae4d9be5c`) and not yet of `main` — which is why its `Closes #2921` trailer never fired: GitHub only auto-closes from the default branch. Checked against `origin/dev` today, not against the commit message:

| # | Problem in this issue | State on `origin/dev` |
|---|---|---|
| 1 | Overwrite guard rests on the reader's `existingNames[]` | Author prompt (`workfly.js:198`): "If a file already exists at `<target>`, write nothing at all"; the script treats an empty `draft.path` as a failure (`workfly.js:277-279`: "The author wrote nothing: … already exists, so the draft was refused rather than overwritten") |
[...]

Issue #2947, opened 2026-09-18
genie update --dev reads the channel manifest from the raw CDN, so a just-published release is invisible for minutes


namastex888:
## Problem

`genie update --dev` resolves the channel manifest through `raw.githubusercontent.com`, whose copy lags the repository by minutes. For that window the command reports the PREVIOUS release as the newest one and installs it, while the Release object and the GitHub API already carry the new build.

It is not a correctness failure — whatever it installs is verified normally — but during a release check it reads as "the release is broken", and an operator who runs the command once after a publish lands a version behind and has no signal that they did.

## Reproduced twice, two different releases

**v5.260918.1 (2026-09-18 ~01:10 UTC)** — release published and `gh api repos/automagik-dev/genie/contents/.well-known/dev.json` decoded to `5.260918.1`, while the raw copy still served `5.260917.1`:

```
$ genie update --dev -y
▸ Channel: dev (dev)
▸ Update available: 5.260916.8 → 5.260917.1     # the previous release
```

The raw copy caught up ~4 minutes later; a second run installed 5.260918.1.

**v5.260918.2 (2026-09-18 ~04:00 UTC)** — same split, observed directly:

```
$ gh api repos/automagik-dev/genie/contents/.well-known/dev.json --jq .content | base64 -d | jq -r .version
5.260918.2
$ curl -s https://raw.githubusercontent.com/automagik-dev/genie/main/.well-known/dev.json | jq -r .version
5.260918.1
```

A wait loop on the raw copy took ~5 minutes to flip, after which the update installed 5.260918.2 normally.

Note the Release object appearing is NOT the signal either: v5.260918.2 existed as a draft with all 20 assets while `publishedAt` was `0001-01-01T00:00:00Z`, and the manifest commit lands later still.

## Where it lives

- `src/genie-commands/update.ts:103` — `const RAW_BASE_URL = 'https://raw.githubusercontent.com';` is the only manifest source.
- `src/genie-commands/update.ts:499` — `fetchLatestManifest` is the single entry point; it resolves the URL then calls `opts.fetcher ?? defaultManifestFetcher`.
- `src/genie-commands/update.ts:549` — the default fetcher is one `curl -fsSL --max-time 5` against that single URL. There is no second attempt and no staleness check.
- `src/genie-commands/__tests__/update.test.ts` — the `fetchLatestManifest (G5)` suite pins every branch of that function.

## Suggested shape (not a decision)

[...]

Issue #2927, opened 2026-09-16
update/doctor are blind to pre-record genie skill dirs (genie-review, pm, wizard, codex-fallback txn) left in agent homes


namastex888:
## Problem
`genie update` retires only skills named in the PREVIOUS install record: `retirementTargets` (`src/lib/skills-installer.ts:1253-1267`) computes `removed = previous.inventory − delivered inventory` and walks `previous.agentDirs × removed`. `genie doctor` compares the record's `inventory` and `agentDirs` against disk and never looks at anything else in a home. So a genie-branded skill dir that predates the record — installed by the plugin/`--all` era before `skills-install.json` existed — is invisible to both, forever.

Observed on the dogfood host on 2026-09-16, after a clean `genie update --stable` to v5.260916.6 whose transcript said `nothing to retire` and whose doctor read `skills: agent dirs — 10/10 recorded homes complete`:

| Path | mtime | What |
|---|---|---|
| `~/.agents/skills/genie-review/` | 2026-07-10 | the pre-consolidation review skill (`name: genie-review`), the one issue #2784 cited |
| `~/.agents/skills/pm/` | 2026-07-10 | dropped in #2805 |
| `~/.agents/skills/wizard/` | 2026-07-10 | dropped by `b0e77cd15` |
| `~/.agents/skills/.genie-codex-fallback-retirement/txn-…/` | 2026-08-17 | a transaction dir (journal.json + quarantined `review` copy) from the agent-sync runtime deleted in `699a48bbd` |

All four were removed by hand, backup-first, into `<GENIE_HOME>/state-backups/manual-skills-retirement-2026-09-16T17-14-48Z/`. Every other server that ran genie before the install record existed will carry the same residue, and `genie update` on it will report `nothing to retire`.

## Ask
1. `genie doctor`: for each recorded `agentDir`, warn on any entry whose `SKILL.md` frontmatter `name` is a skill genie once shipped (the historical roster is derivable: every `skills/*/SKILL.md` ever in git, or a shipped constant) but that is absent from the current inventory, plus the `.genie-codex-fallback-retirement` marker dir. Warn-level, read-only, naming the path.
2. `genie update`: extend the retirement pass (or the legacy-integration retirement, which already owns plugin-era assets ≥ 5.260711.6) to move those proven-genie dirs into the same `skills-retirement-<ts>` backup root, with the same `preserved` receipt on any content mismatch.
[...]

PR #2982, opened 2026-09-18
feat(mikro): script-verified status on every issue-triage answer

namastex888:
## What / why

Tonight's triage round found three issues that were **already fixed on the tree** — but that
verdict lived only in `summary` prose, so nothing could measure it, check it, or act on it.
This adds a typed, additive, **script-verified** `status` field to `issue-triage`.

## D1 — schema + prompt

`IssueTriage.status = { state: 'real' | 'fixed-on-tree' | 'fixed-by-open-pr' | 'unclear',
evidence: [{ kind: 'commit' | 'pr', ref, path?, line? }], downgraded? }`, every field with a
zod default, so an answer that omits `status` entirely parses to
`{ state: 'unclear', evidence: [] }` — the previous behaviour, typed. Existing records and the
minimal record in `call.test.ts` keep parsing unchanged (a test pins that).
`.mikro/agents/issue-triage/SYSTEM.md` gets the field in its Output example plus six lines of
prose: `fixed-on-tree` only with a commit seen in this tree's own log AND a `path:line` that
shows the fix; `fixed-by-open-pr` with a PR number; `real` when the defect is still there;
otherwise `unclear`.

## D2 — the fail-closed rule (`scripts/mikro/status.ts`, new)

A `fixed-on-tree` verdict survives **only if**

1. it names at least one `kind: 'commit'` evidence,
2. **every** commit ref passes `merge-base --is-ancestor <ref> HEAD` in the analysed `--dir`
   (hex object names only — `HEAD`, `--help` or a sentence never reaches the subprocess), and
3. at least one evidence carries `path` + `line` and that `path:line` passes the **same**
   `verifyCitations` check every other citation in the answer passes.

Anything else — an unknown sha, a side-branch commit, a shallow clone, no repository at all, a
missing path, a line past EOF — **fails closed**: the state is rewritten to `unclear` and the
answer gains `status.downgraded: { from: 'fixed-on-tree', reason: '<one line>' }`.

This is a **silent post-validation rewrite**, same family as `applyResolutions` — never a
validation error, never a retry (a retry costs money and invites a fabricated ref). The rule is
a pure function; the runner and the citation check are injected. `call.ts` wires it next to the
citation gate and the per-attempt ledger row gains an additive `status: { state, downgraded? }`.

**`fixed-by-open-pr` is NOT verified script-side** in this slice — there is no `gh` inside the
bwrap boundary — so it passes through as answered.

One premise correction worth recording: the brief suggested routing the ancestor probe through
the boundary session "like the facts calls do". Those calls do **not** go through the boundary —
`prepareFacts` runs on the host *before* `openBoundary`, and so does the whole citation gate
(`isTracked`, `everExisted`). The boundary contains the model's runtime, not the orchestrator's
own verification. The ancestor probe therefore runs on the host exactly where `everExisted`'s
`rev-list` already runs — same trust domain, same fail-closed direction.

## Live proof (2 calls, $0.04 total)

| prompt | `status.state` | verdict | cost | seconds |
|---|---|---|---|---|
| `Triage issue #2921` | `fixed-on-tree` | **survived** — evidence `bebf0991b` is an ancestor of HEAD and `.claude/workflows/workfly.js:198` verifies (the file is 368 lines) | $0.02 | 59.8 |
| `Triage issue #2925` | `real` | passed through untouched (only `fixed-on-tree` is the script's business) | $0.02 | 58.4 |

Both: one attempt, zero errors, zero failed citations. The #2921 ledger row carries
`"status":{"state":"fixed-on-tree"}`.

## Tests (`scripts/mikro/status.test.ts`, 16 tests, real repositories in a tmpdir)

ancestor commit + valid `path:line` survives (and is not even re-allocated) · short ref + a
second row carrying the `path:line` survives · a commit that exists but is **not** an ancestor
of HEAD (real side branch) downgrades · one good ref does not carry a bad one · an unknown sha
downgrades · a ref that is not an object name never reaches the subprocess · no commit evidence
downgrades · a commit with no `path:line` downgrades · `path:line` past EOF downgrades · a path
not in the tree downgrades · a dir that is not a checkout fails closed · `real` / `unclear` /
`fixed-by-open-pr` pass through untouched · an answer with no `status` is returned as it came ·
an omitted `status` parses to the `unclear` default · an unknown state is a schema failure ·
the ledger row carries the final state and the downgrade receipt.

**Watched it fail for the right reason:** mutating rule (b) to `commits.find(() => false)`
(the ancestor check never refutes) failed exactly the 6 ancestor-dependent tests — side branch,
mixed refs, unknown sha, non-object ref, non-repository dir, and the ledger receipt — and
nothing else. Reverted, 16/16 pass.

## Validation

- `bun test scripts/mikro` → **175 pass, 0 fail** (9 files)
- `bunx tsc --noEmit` → clean (exit 0)
- `bunx biome check scripts/mikro .mikro` → clean
- `bun run check` → **2843 pass, 1 skip, 0 fail** across 141 files. The 2 biome warnings are
  the pre-existing `noExcessiveCognitiveComplexity` ones in `src/genie-commands/doctor.ts` and
  `src/lib/orca-orchestration-adapter.ts`, untouched here.

## Explicitly NOT in this slice

- **Open-PR verification** — `fixed-by-open-pr` is taken as answered; no `gh` inside the boundary.
- **A score/bench bar for `status`** — there is no labelled truth for it in the fixtures yet, so
  `score.ts`, `bench.ts` and the fixture truth records are untouched.
- **The history measurement** — re-triaging past issues to measure how often `fixed-on-tree` is
  claimed and how often it survives is a later slice.
- `triage.ts`, `.github/`, `package.json` and every other agent's SYSTEM.md are untouched.


namastex888:
Master-session read of this head (`7571e87ae`), after `review-prep` (mikro, $0.05, 10 claims, 7 flags):

- **The rule, read line by line in `status.ts`:** `fixed-on-tree` survives only with ≥ 1 commit evidence, EVERY commit ref hex-shaped (`^[0-9a-f]{7,40}$` — so a ref can never be parsed by git as an option) and `merge-base --is-ancestor <ref> HEAD` exit 0, and ≥ 1 path:line that the existing citation gate accepts. Every other outcome — unknown object, ambiguous short sha, non-repo dir, spawn failure — rewrites to `unclear` with `downgraded{from,reason}`. No retry, no validation error. That is the brief.
- **Interplay with #2972 (facts inside the bwrap boundary), for whoever merges second:** on THIS base the facts pass and the citation gate run on the host before `openBoundary`, so the worker put the ancestor probe beside them — correct here. #2972 moves facts' git calls into the boundary session through an injected runner; `makeAncestorCheck(dir, run)` already takes a `GitRun`, so the follow-up is one argument at the call site, not a redesign.
- **Flags kept:** the `call.ts` wiring (`statusGate`, the two `verifyStatus` call sites, the ledger key) has no test of its own — the unit tests build their own gate; only `fixed-on-tree` is verified (a wrong `real` is not catchable script-side); `fixed-by-open-pr` is unverified in this slice by design.
- **Live:** #2921 → `fixed-on-tree`, survived (`bebf0991b` is an ancestor; path:line real). #2925 → `real`. $0.04.
- Measurement note: the labelled history set this field was meant to be measured on is **n = 7** over the last 200 merged PRs (this repo rarely closes issues by keyword) — a smoke test, not a rate. The sweep stays manual; labels accumulate forward.

namastex888:
History smoke of this branch's `status` field (master session, $0.38, 14 flash calls): the 7 PRs merged into dev that close an issue by keyword × {merge tree, first-parent tree}, issue NUMBER withheld and title+body frozen in the prompt (the live page and `Closes #n` trailers would leak the label), `--dir` = a detached worktree at the historic sha, this branch's `call.ts`.

| issue | merge tree (truth: fixed) | parent tree (truth: not fixed) |
|---|---|---|
| #2921 | `fixed-on-tree` ✓ | `real` ✓ |
| #2927 | `fixed-on-tree` ✓ | `real` ✓ |
| #2732 | `fixed-on-tree` ✓ | `real` ✓ |
| #2724 | `fixed-on-tree` ✓ | run failed (REPL 30 s timeout, then no JSON) |
| #2669 | run failed (see below) | `real` ✓ |
| #2947 | refused, $0 | refused, $0 |
| ~~#2594~~ | excluded — it is a PR, not an issue (my set builder's defect, fixed) | |

- **Answered runs: 4/4 correct at the merge tree, 4/4 correct at the parent tree, 0 false "fixed", 0 downgrades.** n = 4 per side is a smoke test, not a rate (0/4 bounds a rate below ≈ 50 %).
- **One finding for this PR:** #2669-merge lost an attempt to `schema: status.evidence.2.kind — Invalid enum value. Expected 'commit' | 'pr', received 'test'`. The model invented a kind and the strict enum turned a good answer into a retry (which then hit the REPL timeout). Fail-closed alternative worth a follow-up: drop an evidence row of unknown kind instead of failing the parse — the gate already refuses a verdict left without commit evidence.
- #2947's pair was refused before any model call: the historic tree carries a `.mikro/mikro.yaml` that differs from the trusted one. The config-parity guard doing its job; it also means history measurements only work on trees that predate `.mikro/` or match it.

namastex888:
Follow-up pushed as `5a928daa2` (master session): an evidence row whose `kind` is neither `commit` nor `pr` is dropped before validation instead of failing the parse — the observed `kind: 'test'` case. New test pins the drop AND that the gate then downgrades a `fixed-on-tree` left without commit evidence. `bun test scripts/mikro` 176/0, tsc clean, biome clean.

Issue #2950, opened 2026-09-18
install.sh still reads the channel manifest from the raw CDN only, so a fresh install can land a release behind


namastex888:
## Problem

`install.sh` resolves the channel manifest from `raw.githubusercontent.com` only. #2947 fixed that for `genie update` — the CLI now reads the CDN and the public contents API and takes the newer answer — but a **fresh install** still goes through the CDN alone, so for minutes after a release publishes, `install.sh` installs the previous version with nothing saying so.

Found by the independent review of PR #2948 (F5), which noted the fix landed on one of the two entry points.

## Where

- `install.sh:28` — the single manifest source.
- The CLI-side shape to mirror: `manifestUrlForChannel` / `manifestApiUrlForChannel`, `newerManifest`, and the `Accept: application/vnd.github.raw` header that keeps the bytes the file's own bytes (`src/genie-commands/update.ts`).

## Constraints any fix inherits

- **No credential.** `install.sh` must keep working for an unauthenticated user; the API read is a second anonymous GET whose failure degrades to the CDN answer, never a requirement.
- **The API is cached too** — measured 2026-09-18: CDN `max-age=300`, API `max-age=60, s-maxage=60`. This narrows the window roughly five-fold; it does not close it.
- The API's unauthenticated budget is 60 requests/hour **per IP**, which on CI or a NAT'd network can be spent by other traffic. A rate-limited answer must be indistinguishable from "no answer" and must not be parsed as a manifest.

## Not urgent

A fresh install landing one release behind self-corrects on the next `genie update`, which now reads both sources. This is a consistency gap between the two entry points, not a broken install.

namastex888:
Fixed on dev by #2976 (install.sh reads both manifest sources, newest answer wins, API read bounded with --max-time 5); the duplicate #2983 was closed as superseded and its extra coverage landed as test-only PR #2984.

Issue #2926, opened 2026-09-16
darwin: 6 tests in bun run check fail on macOS while ubuntu CI is green


namastex888:
## Problem
`bun run check` on a darwin host at dev `1b6ae7e6d` (Bun 1.3.14, macOS 25.5) fails 6 tests that pass on ubuntu CI. None is touched by recent PRs; they are host-environment failures the ubuntu-only CI cannot see:

- `src/lib/orca-orchestration-adapter.test.ts:269` — expects executable `orca-ide`, gets `orca`: the assertion hardcodes the linux resolution while `resolveOrcaExecutable` returns `orca` on darwin (`src/lib/orca-orchestration-adapter.ts:354`).
- `src/genie-commands/doctor.test.ts` — `--fix-global-db backs up a database with a live WAL completely`.
- `src/genie-commands/__tests__/update.test.ts` — `aborts when the signed release name is for another platform, without a credential to hide behind`.
- `scripts/release-archive-safety.test.ts` — `an escaping member name is rejected before extraction`.
- `scripts/skills-retirement-restore.test.ts` — `documented command 0 restores the tree and leaves existing modes alone`.
- `scripts/design-review-evidence.test.ts` — `the wish copy runs with no sibling skill installed`.

Also, `bun run typecheck` fails on a checkout whose root `node_modules` predates the `react-dom` devDependency (`Cannot find module 'react-dom/server'` from `plugins/dsh-genie-board/src/client.test.ts`) until `bun install --frozen-lockfile` is re-run; a note in the check script or CLAUDE.md would save the next dogfooder the detour.

## Ask
Pin the platform in the orca assertion, and either add a darwin runner to CI or mark the five remaining tests with the host condition they need. Full log available from the 2026-09-16 dogfood session.

namastex888:
Re-verified 2026-09-19: the design-review and update tests were fixed via #2941 and later work, but the assertion this issue cites — orca-orchestration-adapter.test.ts:271 `executable: 'orca-ide'` fed by the platform-resolving default adapter — is unchanged and still fails on darwin; the other four tests have no commits since. There is still no darwin CI runner. Stable is declared for Linux and macOS, so this stays open as a v6 stable blocker.

namastex888:
[...]

PR #2983, opened 2026-09-18
fix(install): read both manifest sources, newest answer wins

namastex888:
Closes #2950 — the install-side twin of #2947. `genie update` learned to read both manifest sources in #2948/#2951; `install.sh` kept the single one, so a **fresh** install still landed a version behind for minutes after a release published, silently.

## The change

Both sources publish the same file and neither needs a credential, so both are read and the newer answer wins.

| Piece | What it does |
|---|---|
| `manifest_file_name` | shared by both URL builders, so the two can never disagree about which file a channel means |
| `resolve_manifest_api_url` | `…/contents/.well-known/<file>?ref=main`, asked with `Accept: application/vnd.github.raw` so the response is the file's own bytes |
| `version_is_newer` | compares only plain dotted numbers; anything else answers false, so an unparseable or prereleased version never displaces the CDN. A tie keeps the CDN. |
| `manifest_is_usable` | requires the channel to match **and** a version to be present |

The rules are deliberately conservative: the failure being prevented (a fresh install one release behind, which self-corrects on the next `genie update`) is much smaller than the failure a wrong answer would cause.

## What cannot win

A rate-limit body, an error page, and the API's JSON envelope all lack `channel`, so none can displace a stale-but-valid CDN answer — **not even when the envelope's inner version is higher**, which the test asserts with `9.260918.9` hidden inside it.

Both pre-existing failure contracts survive: **exit 5** naming the CDN url when neither source answers, **exit 1** on channel mismatch.

## Honesty about the window

Measured today: CDN `max-age=300`, API `max-age=60, s-maxage=60`. This narrows the stale window about fivefold; it does not close it. The comment in the file says that rather than claiming the API is immediate — the wrong version of that sentence was a finding against the CLI-side fix (#2951, F1).

## Validation

- `bun run check` on darwin: **2684 pass, 1 skip, 6 fail** — the six #2926 darwin names, unchanged.
- Five new tests in `scripts/install-manifest.test.ts` run the **real** `fetch_latest` with `curl` shimmed onto `PATH`: no case touches the network. The shim records its argv, so the raw `Accept` header is asserted **from what was actually sent** rather than from the shape of the source file — the weakness the #2948 review flagged in the CLI twin, avoided here from the start.
- Mutation-checked: force the CDN to win → the stale-CDN test fails; drop the `Accept` header → the header test fails. Each kills exactly one test.
- `version_is_newer` exercised across ties, a `10` vs `9` component, a major-version jump, a non-numeric value and an `-rc1` suffix.

## Note for the reviewer

`install.sh` runs `main` when sourced unless `GENIE_INSTALL_SOURCE_ONLY=1` is set. I learned that the hard way while building this — sourcing it without the guard performed a real stable-channel install on my host and downgraded it. The new tests set the guard; anything you write that sources the script should too.


namastex888:
Closing in favour of #2984, which keeps this PR's test file and drops its `install.sh` change as duplicated work.

`dev` already carries the same fix: #2976 merged as `81aab53e2`, "fix(install): read the channel manifest from the CDN and the contents API". The two were written independently overnight against the same issue (#2950). Line by line:

- `MANIFEST_API_BASE="https://api.github.com/repos/${REPO}/contents/.well-known"` — identical in both.
- `manifest_file_name` (here) / `manifest_file_for_channel` (dev) — same body, same `stable→latest.json` rule, different name; dev's `resolve_manifest_url` and `resolve_manifest_api_url` are shared through it exactly as here.
- `resolve_manifest_api_url` — both emit `<api base>/<file>?ref=main`.
- `Accept: application/vnd.github.raw` on the API read — both.
- "Newest answer wins": dev's `version_is_newer` compares **every** dotted component in a loop and fails closed on any non-integer component; this PR's compares three fixed components via `10#`. dev's is a superset — `5.260918.10` vs `5.260918.9` is pinned by its own test.
- Tie keeps the CDN bytes — both.
- CDN silent → install from the API — both (dev additionally `warn`s).
- API failed, rate-limited, or answering with a base64 envelope → keep the CDN — both; dev discards the API payload through the same `manifest_channel_matches` check.
- Both sources gone → `die "could not fetch $url" 5` — both.
- dev additionally bounds the optional API read with `--max-time 5`, which this PR does not, so a hung API can stall an install the CDN would have served.

The one behaviour here that dev does not have: when the CDN answers for the **wrong channel** and the API answers correctly, this PR installs from the API, whereas dev still exits 1 with `manifest channel mismatch`. That relaxes a pre-existing safety contract rather than adding capability — a channel mismatch on the published CDN copy is a producer-side bug worth failing on — so it was deliberately not carried over.

What was worth keeping is `scripts/install-manifest.test.ts`: it covers five cases dev's own `install-swap.test.ts` leaves open, two of which are the behaviours the fix actually added (the API answering alone, and the request shapes asserted from recorded `curl` argv). All of them pass against dev's `install.sh` unmodified, and the two new ones fail against `81aab53e2~1` for the right reason. #2984 carries that file, minus its one test that duplicated an existing one.

Nothing is lost by closing this.

PR #3007, opened 2026-09-19
fix(release-orphan-alert): auto-close healed incidents and skip promotion tags

namastex888:
## What

`.github/workflows/release-orphan-alert.yml` now reconciles both directions on every cron fire, and stops alarming on the designed human gate.

**Auto-close.** The releases query is widened to `--jq '.[] | [.tag_name, .html_url] | @tsv'`. Every OPEN issue carrying the `release-incident` label whose tag now has a Release gets the Release URL commented as the evidence for the close, then gets closed (`gh issue comment` → `gh issue close --reason completed`). The pass runs before the orphan loop's early `exit 0`, so a fire with zero orphans still reconciles. Closed incidents are never touched — a closed issue is still an operator ack.

**Promotion-tag skip.** version.yml mints a `v*` tag on exactly two paths: on **dev** it commits the version-only child `chore(version): bump to <version> [auto-version]` on top of the CI head, pushes it with the tag in one `git push --atomic`, and calls release.yml itself (`release_ready=true`); on a verified **main promotion** it tags the promotion HEAD as it stands (`git tag "v${VERSION}" HEAD`), pushes the tag alone, and sets `channel=stable` / `release_ready=false` because stable must be started by a maintainer and approved by a different one. A tag whose own commit is not that `[auto-version]` child was minted on the promotion path, so it legitimately has no Release and never reaches `gh issue create`.

**Unchanged:** the exact-title `--state all` dedup and its fail-closed `MATCH_STATUS` split, the `[30min, 24h]` constants byte-identical, `permissions: contents: read` + `issues: write`, the SHA-pinned checkout. No release or tag mutation verb is reachable.

## Evidence

`scripts/release-orphan-alert.test.ts` is new and reads the workflow file.

Against **origin/dev's** workflow (no `gh issue close` exists today) it fails as designed:

```
6 pass / 9 fail
(fail) lists only OPEN issues carrying the release-incident label
(fail) closes each healed incident
(fail) comments the Release URL before closing
(fail) resolves the Release URL from a releases query that yields tag AND html_url
(fail) runs before the orphan loop short-circuits on zero orphans
(fail) the new listing fails loud — no swallowed API error
(fail) skips a tag whose commit is not the [auto-version] version child
(fail) the predicate guards gh issue create (the skip precedes it and continues)
(fail) a comment adjacent to the predicate names both of version.yml mint paths
```

Against this branch: `15 pass / 0 fail`.

| Command | Exit |
|---|---|
| `bun test scripts/release-orphan-alert.test.ts scripts/release-docs.test.ts` (58 pass / 0 fail) | 0 |
| `bun run lint` (2 pre-existing complexity warnings only) | 0 |
| `bun run typecheck` | 0 |
| `Bun.YAML.parse` of the workflow + `bash -n` on the extracted `run:` block | 0 |

The embedded script was also executed end to end against a throwaway git repo and a stub `gh`: it closed the one healed incident with the URL comment first, left a label-carrying issue with a foreign title open, skipped the `Merge pull request …/dev` promotion tag, and filed only the genuine `[auto-version]` orphans. An empty open-incident listing is a clean no-op; a `gh` failure aborts the job (exit 4).

## Not touched

`docs/_internal/runbooks/release-pipeline.md` — `docs/` is the `.docs-vendor` submodule and `git submodule status` reports it with a `-` prefix (not checked out), so the runbook page is a separate change in automagik-dev/docs.

Part of wish `v6-stable-cut` G2; executed by hand because the change touches `.github/`.


namastex888:
## FIX-FIRST repairs applied — `b62b8d5259a71f466f256bc81080af5a28fcab3f`

| Item | Fix | Pinned by |
|---|---|---|
| HIGH SIGPIPE | `RELEASE_URL=$(awk -F'\t' -v tag="$INCIDENT_TAG" '$1 == tag { print $2; exit }' <<<"$RELEASES_TSV")`; `RELEASE_TAGS` uses a herestring too | `the Release-URL lookup is a herestring, never a pipeline into an exiting awk` (also bans `printf … \|` and `\| awk/cut` on any executable line) |
| MEDIUM drafts | `--jq '.[] \| select(.draft == false) \| [.tag_name, .html_url] \| @tsv'` — both the close pass and the has-release set read the filtered rows | `drafts are not Releases …` + `the jq program actually drops a draft release` (extracts the jq program from the workflow and runs it) |
| MEDIUM reopen | one shot per issue: `gh issue view --json comments --jq '[… select(startswith($RESOLUTION_PREFIX))] \| length'` > 0 → skip; the comment body starts with that exact prefix | `auto-closes an issue at most once, so a deliberate reopen stands` |
| MEDIUM limit | `--limit 1000` | `the open-incident listing is not truncated at 100` |
| LOW | `::notice::` on the promotion skip; empty `git log` subject → `::warning::` + skip the check (never a promotion verdict); `</dev/null` on all three `gh` calls in the loop | `the skip is an annotation, and an unreadable subject is a warning…` + `every gh call inside the read loop is detached from the loop stdin` |

All seven new assertions fail against the previous commit `412c724fd` (15 pass / 7 fail) and pass here.

| Command | Exit |
|---|---|
| `bun test scripts/release-orphan-alert.test.ts scripts/release-docs.test.ts` (65 pass / 0 fail) | 0 |
| `shellcheck -s bash` on the extracted `run:` block | 0, clean |
| `bash -n` on the same block + `Bun.YAML.parse` | 0 |
| `bun run lint` (2 pre-existing complexity warnings) / `bun run typecheck` | 0 / 0 |

Re-executed end to end against a throwaway repo and a stub `gh` with a **200 KB** releases TSV: rc=0 (no 141), the healed incident closed with the URL comment first, an already-auto-resolved (deliberately reopened) issue left open, the `Merge pull request …/dev` promotion tag skipped with a `::notice::`, and only the genuine `[auto-version]` orphan filed.

namastex888:
## Round-2 repair applied — `28039644e7abbb33ed70451d61696fe50641388f`

The one-shot receipt is no longer the resolution comment. It is the label **`release-auto-resolved`**, written **after** the close and read from the listing this pass already fetches (`--json number,title,labels`).

Order is now **comment (evidence) → close (the outcome) → `gh issue edit --add-label`(the receipt)**, so each failure mode degrades to something recoverable:

| Failure | Before | Now |
|---|---|---|
| close fails after the comment posted | issue open **with** the receipt → skipped as "reopened deliberately" **forever** | no receipt → next fire retries; one duplicate comment at worst |
| label edit fails | n/a | issue is already closed (the outcome that mattered); next fire re-closes at worst |
| operator reopens deliberately | honoured | honoured — the label survives the reopen |

The label is ensured once per step with `gh label create "$RESOLVED_LABEL" --repo "$REPO" --force …` (idempotent, covered by the existing `issues: write`). The per-incident `gh issue view` and the jq interpolation of `RESOLUTION_PREFIX` are deleted — one fewer API call per issue, and no shell string interpolated into a jq program.

**Backlog guard (LOW):** after the increment, `[[ "$CLOSED_COUNT" -ge 50 ]] && break` with a `::notice::`. GITHUB_TOKEN allows 1,000 REST writes/hour/repo; 664 incidents × 3 writes would burn the budget in one fire and take the orphan filing down with it.

### Evidence

Three changed/new assertions fail against `b62b8d5` (21 pass / 3 fail) and pass here: `auto-closes an issue at most once…` (now pins the label path and asserts `gh issue view` / `RESOLUTION_PREFIX` are gone), `the receipt label is written AFTER the close…` (comment < close < label edit), `the close pass is bounded per fire…`.

Live re-simulation against a stub `gh`:

```
FIRE 1 (comment ok, close FAILS)  -> exit 1, receipt label present after fire 1: NONE
FIRE 2 (close succeeds)           -> STUB CLOSE ok / STUB LABEL ADDED; auto-closed 1
FIRE 3 (receipt present)          -> "already auto-resolved once (reopened deliberately); leaving open", exit 0
60-incident backlog               -> "::notice::auto-close budget reached (50 issues this fire)", auto-closed 50
```

| Command | Exit |
|---|---|
| `bun test scripts/release-orphan-alert.test.ts scripts/release-docs.test.ts` (67 pass / 0 fail) | 0 |
| `shellcheck -s bash` on the extracted `run:` block | 0, clean |
| `bash -n` + `Bun.YAML.parse` | 0 |
| `bun run lint` / `bun run typecheck` | 0 / 0 |

Note on the red check: `Unit (darwin)` failed on the macOS-only flake `timeline verbs under a concurrent delete > report never leaks a raw FOREIGN KEY failure` (`src/term-commands/v5-task.test.ts`), which passes locally 99/99 and is unreachable from a change confined to `.github/workflows/` and `scripts/`. The failed jobs were re-run.

namastex888:
## Round-3 fix — `082163990b2c720250dd73b9029f49ffcdf7f75a`

`gh label create … --force` at the top of the step is replaced by `ensure_resolved_label`, called immediately before the first `gh issue edit --add-label`:

- **No `--force`.** It rewrites an existing label's description and colour on every fire, overriding an operator who adjusted either.
- **Lazy.** The label API is touched only when an incident is actually being closed; a fire that closes nothing makes no label call at all.
- **Only when absent.** `gh label list --repo "$REPO" --limit 200 --json name --jq '.[].name' </dev/null` into a variable, then a fail-closed `grep -Fxq … || lookup_status=$?` — only status 1 means absent, anything else aborts (same posture as the issue-title dedup). No `|| true`.
- **Memoized** with `RESOLVED_LABEL_READY=1`: fifty closes cost one label read and at most one write.

One deviation from the suggested snippet, deliberately: the check reads the listing into a variable and greps a herestring instead of `gh label list … | grep -Fxq`. `grep -q` leaves the pipe early, so under `pipefail` a successful match becomes exit 141 once the label list is large — the exact SIGPIPE class removed from the Release-URL lookup in round 2.

### Evidence

Both changed assertions fail against `2803964` (23 pass / 2 fail) and pass here: `the receipt label is created lazily, only when absent, and never with --force` (no `--force` in the create call; exactly one `gh label create` in the file, inside the helper, after the open-incident listing; `ensure_resolved_label` is the line immediately above the `--add-label` edit) and the updated `auto-closes an issue at most once…`.

Stub-`gh` simulation:

```
label ABSENT  -> 2 closes, ONE create (argv carries no --force), 2 --add-label edits
label PRESENT -> 0 creates
closes nothing -> neither label verb called
```

| Command | Exit |
|---|---|
| `bun test scripts/release-orphan-alert.test.ts scripts/release-docs.test.ts` (68 pass / 0 fail) | 0 |
| `shellcheck -s bash` on the extracted `run:` block | 0, clean |
| `bash -n` + `Bun.YAML.parse` | 0 |
| `bun run lint` / `bun run typecheck` | 0 / 0 |

PR #3015, opened 2026-09-19
feat(workflow-loader): ship it in the payload, and prove the row boots

namastex888:
## What this is

[#3011](https://github.com/automagik-dev/genie/pull/3011) landed the loader as source only: the tarball did not carry it and no host had ever loaded its row. This closes both halves — the payload wiring, and a smoke that boots a real Host with the row active.

## Payload wiring

- `scripts/build-binary.sh` builds the loader's dist into the stage and requires its `package.json`, `agent.cordis.yml`, `cordis.patch.yml`, `README.md` and `dist/index.js` — the same explicit member discipline the board's row already follows.
- `scripts/release-payload-version.ts` gains `plugins/dsh-workflow-loader/package.json` in `TOP_LEVEL_VERSION_FILES`, so the release version is stamped into it and verified against `VERSION`. Its test fixture carries the file, so a payload missing it now fails closed **by name**.

## The smoke, and what it caught

`scripts/dsh-workflow-loader-smoke.ts` follows the board smoke's harness: build (refusing a stale dist first), install into an isolated `DSH_HOME`/`HOME`/`TMPDIR` tree, assert the composed profile carries the row, boot a real `dsh web`, exchange the launch token for a cookie, then hold the Host through the same 10 s settle window the board smoke documents — *a printed URL is not a boot*, and `N entries did not activate` is the failure that window exists for.

**It earned its keep on the first run: the row failed to load, twice.**

1. The shipped `cordis.patch.yml` spells `journalDir: ''` for "not configured", and the resolver rejected an empty string — so the patch's own default made the Host refuse the row. The key is now commented out (the documented default applies) and an empty string is accepted. Caught with an explicit `ValidationError: invalid config: expected a non-empty path (at journalDir)`.
2. `cordis` validates a patch through the row's `validate()`, which returns the **resolved** config, and then hands that object to `apply()`, which resolved again. The resolver rejected its own output (`userRoot: ''`) and the boot **hung with no output at all** — no error, no URL, 45 s of silence. The resolver is now idempotent (the property that was missing) and a test pins it: `resolveLoaderConfig(resolveLoaderConfig(x))` must equal `resolveLoaderConfig(x)` for every shape the row can emit.

Without both fixes the row would have taken a profile down on its next restart — the same class of failure this repository records as a Host killed by three pending entries. I would not have installed it anywhere before this ran.

## Evidence

```
$ bun scripts/dsh-workflow-loader-smoke.ts
bun run build: OK
dsh plugin --profile web add link:...: OK
dsh plugin --profile web list --depth 0: OK
dsh --profile web --dump-config: OK
composed profile carries genie-dsh-workflow-loader: OK
loader row activated and the Host stayed up for 10000 ms: OK
dsh-workflow-loader smoke: PASS
```

- 51 tests pass (44 plugin + 7 smoke unit), `biome check` clean on all 14 touched files.
- The stamping change is proven locally — stamp and verify succeed with the loader present, and a payload missing it is refused by name — because `release-payload-version.test.ts` uses `mkdtemp(tmpdir())` and this host's `TMPDIR` is read-only outside CI.

## Not in this PR

- Installing it into a live profile: that is an operator action with a restart, and it is the next step, not this PR's.
- The wish re-plan for `dsh-workflow-fork` and the stale `h1-agent-builder` team: both tracked separately.


namastex888:
## Independent review — **FIX-FIRST** (1 HIGH)

Reviewed commit: `b582fe7ce` ("ship it in the payload, and prove the row boots"), detached worktree, deps installed.

### HIGH

1. **`scripts/verify-dsh-genie-board-release.test.ts:29-34` — the other payload fixture was missed; 20 tests now fail.**
   This PR adds `plugins/dsh-workflow-loader/package.json` to `TOP_LEVEL_VERSION_FILES` (`scripts/release-payload-version.ts:20`), but only `scripts/release-payload-version.test.ts` got the matching fixture entry. Every fixture in `verify-dsh-genie-board-release.test.ts` now throws `release payload metadata is missing: <tmp>/payload/plugins/dsh-workflow-loader/package.json`.
   Evidence: `bun test scripts/verify-dsh-genie-board-release.test.ts` → **5 pass / 20 fail** at this head (reproduced in a fresh detached worktree), and remote CI on this head is red on three checks: `gh pr checks 3015` → Quality Gate, Unit (linux), Unit (darwin).
   **Fix:** add `'plugins/dsh-workflow-loader/package.json'` to the fixture array at lines 29-33.

### MEDIUM

2. **`scripts/dsh-workflow-loader-smoke.ts` is wired into nothing.** No workflow, no package script, no README mention. It never skips either: with no `dsh` on `PATH` it hard-fails, so it cannot be added to CI as-is. Decide the wiring (a CI job that installs `dsh`, or "owner-run only" stated in the file header and README) and make the missing-`dsh` path say so explicitly instead of failing as a boot error.
3. **`scripts/verify-dsh-genie-board-release.ts:16-24` — "single source of truth for the DSH plugin payload members" is now false.** The release verifier checks the board's members only, so the loader's `agent.cordis.yml`, `cordis.patch.yml`, `README.md` and `dist/index.js` are verified at build time only. Extend the verifier's member list or narrow the claim.
4. **`plugins/dsh-workflow-loader/README.md:125-133` — "Not done yet" is stale.** It says the payload does not carry the plugin and that there is no host smoke; both are now true statements in the past tense.
5. **`AGENTS.md:57` (release contract)** names only the board's DSH payload and "both plugin packages"; the payload now carries three. Name the loader.

### LOW

6. Nothing pins the loader entry in `TOP_LEVEL_VERSION_FILES` (`scripts/release-payload-version.test.ts:36`), so removing it later would not fail a test. Assert membership explicitly.
7. The two new `plugins/dsh-workflow-loader/src/tool.test.ts` regressions restate hand-copied literals instead of reading the shipped `cordis.patch.yml`; read the shipped file so the test fails when the patch drifts.

### Verified clean at this head

All five payload members are required fail-closed by name and the require list matches `build.ts` output; stamping reaches the loader manifest and refuses a payload missing it by name; the empty-string `journalDir`/`userRoot` fix is correct, minimal and idempotent (the cordis double-resolve claim is confirmed against cordis 4.0.2 source); `inject: ['tools','workflowEngine']` matches the pre-existing ctx reads; typecheck, biome, knip and `bash -n` are clean.

**Not verified:** a real `dsh web` boot was not re-run; the smoke's logic was judged by reading, not by execution.

**For the fix agent:** the fix is one array entry plus the wiring decisions; run `bun test scripts/verify-dsh-genie-board-release.test.ts scripts/release-payload-version.test.ts plugins/dsh-workflow-loader/src/tool.test.ts` and then the repo gate `bun run check`.

---
<sub>Two models reviewed this diff independently at `b582fe7ce` — `deepseek/deepseek-v4-flash` (xhigh) and `zai/glm-5.3-flash` (high) — in separate detached worktrees; the HIGH above was reproduced a third time in the orchestrating session and corroborated against this PR's own CI runs.</sub>

